PR #23, opened 2023-06-07
Add more information about userland implementations

benlesh:
+ adds information about userland implementations with links
+ adds information about userland support
+ adds bit about `symbol-observable` which helps illustrate a larger number of libraries that care about exactly the same signature.


domfarolino (approved):
I think this looks really great, thanks a lot! My only concern with this is that there are several instances where we say an external library is "nearly" or "mostly" "identical" with this one.

I wonder if this is over-selling the similarities though, since this proposal already deviates from RxJS by using AbortController/Signal and the ctor accepting a dictionary of functions instead of just the functions themselves. Maybe the semantics are what we're really trying to comment on here, in which case "nearly identical" is not an oversell, but I just wanted to draw some last-minute attention to those obvious ABI differences and see if you had any thoughts, or if you think we should just merge this in.

tbondwilkinson:
I think the relevant bit is whether all existing Observable-likes are adaptable to a native Observable with shim layers, rather than that conform exactly to the native API.

E.g. AbortController/Signal is an implementation detail, as is the specific ctor syntax.

domfarolino:
Yeah I think that's a better way of looking at it. I'm not sure how to judge that I suppose, maybe @benlesh has more to add on that? (This PR is probably good to go as-is, I just wanna poke a little since I don't know these other implementations well).

benlesh:
I'll give it some thought, of course.

PR #32, opened 2023-07-28
Correcting typos

appsforartists:
- `done` is called `complete` elsewhere in the spec
- `first` is a method that needs parens


domfarolino (approved):
Nice catch, thanks a lot!

PR #36, opened 2023-07-29
Remove redundant function declaration

adamochayon:
The `finish` method in example 5 is unused. Removing it will make the example easier to quickly grok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,10 +116,6 @@ function streamStock(ticker) {
   });
 }
 
-function finish(stopMsg) {
-  socket.send(JSON.stringify(stopMsg));
-}
-
 const googTrades = streamStock('GOOG');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,10 +116,6 @@ function streamStock(ticker) {
   });
 }
 
-function finish(stopMsg) {
-  socket.send(JSON.stringify(stopMsg));
-}
-
```
domfarolino, line 122: Nice catch. I think the intention was to put `finish` up above on L106, instead of `socket.send(JSON.stringify(stopMsg))`. But this PR (keeping this function's contents inlined up above instead of using it) is probably simpler.

domfarolino (approved):

PR #40, opened 2023-07-31
Fix `forEach` and `toArray` return types

### LINKED ISSUE #27: Return types of some operators seem off
- `toArray()` is specced to return an `Observable`, but per its name, I think it should return an `Array` (like iterator helpers) or a `Promise<sequence<any>>` (like async iterator helpers).

- `forEach()` is specced to return an `Observable`, but I would expect it to either return `undefined` (like `Array.prototype` / iterator helpers) or `Promise<undefined>` (like async iterator helpers)

domfarolino:
Closes #27. Per discussion there, it seems we've landed on:
 - `toArray()` causing a subscription, and asynchronously yielding an array via a Promise, which seems natural
 - `forEach()` per https://github.com/domfarolino/observable/issues/29#issuecomment-1656292907, this also causes a subscription, so it shouldn't return an `Observable`. And since there's no accumulated or passed-through values, it returns undefined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ interface Observable {
   Observable drop(unsigned long long);
   Observable flatMap(Function project);
   Promise<sequence<any>> toArray();
-  undefined forEach(Function callback);
+  Promise<undefined> forEach(Function callback);
 
   // Promise-returning. See "Concerns" section below.
   Promise<any> every(Predicate predicate);
--- a/README.md
+++ b/README.md
@@ -324,8 +324,8 @@ interface Observable {
   Observable take(unsigned long long);
   Observable drop(unsigned long long);
   Observable flatMap(Function project);
-  Observable toArray();
-  Observable forEach(Function callback);
+  Promise<sequence<any>> toArray();
+  undefined forEach(Function callback);
 
   // Promise-returning. See "Concerns" section below.
   Promise<any> every(Predicate predicate);


README.md
```diff
@@ -324,8 +324,8 @@ interface Observable {
   Observable take(unsigned long long);
   Observable drop(unsigned long long);
   Observable flatMap(Function project);
-  Observable toArray();
-  Observable forEach(Function callback);
+  Promise<sequence<any>> toArray();
+  undefined forEach(Function callback);
```
benlesh, line 328: Issue: This should return a `Promise<void>` in TS terms that resolves when the observable completes, or errors if the observable errors.

```suggestion
  Promise<undefined> forEach(Function callback);
```

README.md
```diff
@@ -324,8 +324,8 @@ interface Observable {
   Observable take(unsigned long long);
   Observable drop(unsigned long long);
   Observable flatMap(Function project);
-  Observable toArray();
-  Observable forEach(Function callback);
+  Promise<sequence<any>> toArray();
```
benlesh, line 327: Suggestion: I'd almost recommend we drop this entirely for now. It could be added in a follow up.

benlesh (changes requested):

README.md
domfarolino: Let's consider that outside of this PR. I think initially, it would be nice to match iterator helpers as closely as possible, and `toArray()` is one of those helpers, so if we remove it we should do so with careful public notice that we're deviating.

benlesh: Ah! I had missed it was one of those helpers. I often don't like how it's used in RxJS. Haha.

benlesh (approved):

PR #42, opened 2023-08-01
Create initial reference implementation

benlesh:
Just a spike in TypeScript. Will add tests in a followup, add compilation to JavaScript, etc.


benlesh:
Could even just convert it to JavaScript. I'm relatively indifferent.

impl/observable.ts
```diff
@@ -0,0 +1,496 @@
+/**
+MIT License
```
domfarolino, line 2: Hmm, should we perhaps make this a separate LICENSE file in the top-level of the repo?

benlesh: I think that's fine. Mostly I like licensing things because of the liability.

benlesh:
@domfarolino should we merge this? is it useful?

impl/observable.js
```diff
@@ -0,0 +1,650 @@
...
+				});
+			},
+			[Symbol.asyncIterator]() {
+				return this;
+			},
+		};
+	}
+}
+class Subscriber {
+	#active = true;
```
gitawego, line 553: just a question, do we really need to use private properties ? as it's not widely available in browsers.

bakkot: Private fields have been available in evergreen browsers for more than two years. https://caniuse.com/mdn-javascript_classes_private_class_fields

ljharb: Also, this isn't a polyfill, it's a reference implementation. Polyfills should generally strive for maximal backwards compatibility, but reference implementations should strive for the most modern standard syntax and APIs available.

LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Ben Lesh <ben@benlesh.com>, Domenic Farolino <dom@chromium.org>
```
domfarolino, line 3: Dominic. Not Domenic

domfarolino:
See https://github.com/WICG/observable/pull/42#issuecomment-1689972876.

impl/observable-polyfill.d.ts
```diff
@@ -0,0 +1,379 @@
...
+		on<K extends keyof AbortSignalEventMap>(
+			eventType: K,
+		): Observable<AbortSignalEventMap[K]>;
+	}
+	interface AbstractWorker {
+		on<K extends keyof AbstractWorkerEventMap>(
+			eventType: K,
+		): Observable<AbstractWorkerEventMap[K]>;
+	}
+	interface Animation {
```
domfarolino, line 17: Why are all of these being added?

interfaces.txt
```diff
@@ -0,0 +1,73 @@
+AbortSignal
```
domfarolino, line 1: What is this file?

impl/observable-polyfill.d.ts
benlesh: Honestly? Just to be nice. If anyone wants to use this with TypeScript, these types will help them. They don't *need* to be here.

interfaces.txt
benlesh: Cruft. Sorry.

LICENSE.txt
benlesh: Sorry... 😬

impl/observable-polyfill.d.ts
benlesh: My two cents is we leave them. They can't do anything but help folks use this and try it out.

keithamus: They map the event name strings to the event subclasses. This way `eventTarget.on('click')` will typed as an `Observable<MouseEvent>` while `eventTarget.on('keypress')` will be typed as an `Observable<KeyboardEvent>`.

impl/observable.js
```diff
@@ -0,0 +1,650 @@
...
+	#abortController = new AbortController();
+	#signal;
+	#teardowns;
+	get signal() {
+		return this.#signal;
+	}
+	get isActive() {
+		return this.#active && !this.signal.aborted;
```
keithamus, line 561: This doesn't seem to align to the current spec. AFAICT it's possible to see `isActive === false` while also `this.signal.aborted === false`.

impl/observable.js
```diff
@@ -0,0 +1,650 @@
...
+				? { next: fnOrObserver }
+				: fnOrObserver;
+		const observer = {
+			next: noop,
+			error: reportError,
+			complete: noop,
+			...partialObserver,
+		};
+		const subscriber = new Subscriber(options?.signal, observer);
+		this.start(subscriber);
```
keithamus, line 74: Does this need to try/catch?

domfarolino:
This PR really hasn't seen any activity in months, and with Chromium getting pretty close to shipping Observables, I feel like we should probably just close this PR out. But feel free to re-open if you have plans to get to it.

dariomannu:
Hey, shame this has been closed, as it was a good start.

OK that Chrome has it built-in, but it's still feature-flagged and other browsers don't have it at all.
Any strong opinions against reviving this and having an "official" polyfill? I'm happy to update this to the current spec, package it up, help co-maintain it, create some examples, tests, stackblitzes, whatever...

PR #49, opened 2023-08-29
Fix `finally` to return `Observable`

### LINKED ISSUE #48: `finally` should return `Observable`
Currently it's returning `undefined`.

https://github.com/domfarolino/observable/blob/master/README.md?plain=1#L318

benlesh:
closes #48

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 30, 2024, 2:24 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Fobservable%2F5b136db3304f7c0cb9b6c354a903953d273c75e2%2Fspec.bs&md-warning=not%20ready)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/observable%2349.)._
</details>


README.md
```diff
@@ -315,7 +315,7 @@ interface Observable {
   constructor(SubscribeCallback callback);
   undefined subscribe(Observer observer);
 
-  undefined finally(VoidFunction callback);
+  Observable finally(VoidFunction callback);
```
domfarolino, line 318: Can you move this to the "Observable-returning operators" section below?

domfarolino, line 318: Also, does `finally()` still need the `VoidFunction` callback here? Probably not, right, if we're observing completion via the returned Observable. That brings me to one more question: why return an Observable instead of a Promise?

If we're just trying to understand if the Observable finished in any manner (regardless of whether it did so in error or successfully), like `Promise.prototype.finally()`, maybe this should just be a Promise-returning API?

At the very least, we need to update https://github.com/domfarolino/observable/blob/master/README.md?plain=1#L427-L429 in this PR.

benlesh: Yeah, it needs it. `finally` callback would fire if the subscription ends for any reason: error, complete, or abort/unsubscribe.

domfarolino: Isn't that what we're using the returned Observable for? Why have both? It seems that this should either:
 - Return undefined and take a callback
 - Return an Observable (whose `complete()` callback is called whenever the observable instance that `finally()` was called on ends for any reason) and takes **no** callback

Are both needed?

benlesh: We discussed this offline, and we worked it out. I think we agreed this can be merged, @domfarolino

domfarolino: Maybe we could just add some documentation just above this method mentioning that this is a pass-through for the observable that this is called on (i.e., the returned observable is "identical"), but that the `finally()` handler gets called.... *after*(?) the complete()/error() events of the final observable?

domfarolino:
I think just a tiny bit of documentation to make this slightly more clear (https://github.com/WICG/observable/pull/49#discussion_r1327094796) and resolving the conflicts is all that's needed for this at this point.

benlesh:
@domfarolino updated the documentation about it

README.md
```diff
@@ -368,6 +367,7 @@ interface Observable {
   Observable take(unsigned long long);
   Observable drop(unsigned long long);
   Observable flatMap(Mapper mapper);
+  Observable finally(VoidFunction callback);
```
domfarolino, line 370: Seems that https://github.com/WICG/observable/pull/49#discussion_r1324578891 has not been addressed yet.

benlesh: Updated.

domfarolino: I think I'm still confused why this takes a callback and returns an Observable. That's what I'm pointing to above: https://github.com/WICG/observable/pull/49#discussion_r1324578891.

domfarolino:
It looks like this will need to be rebased on top of the changes that ripped the IDL out of the README and into the spec.

README.md
benlesh: it would technically get called _after_ the error or `error`, `complete`.

```ts
source
  .finally(() => console.log('finalized'))
  .subscribe({
    complete: () => console.log('complete'),
  });
  
// logs
// "complete"
// "finalized"
```

benlesh: Oh... looks like I already did this work... just circling through all of this.

benlesh:
Rebased.

domfarolino:
I can't seem to get your attention on https://github.com/WICG/observable/pull/49#discussion_r1412254718 so let me mention it directly here :)

README.md
benlesh: Ah... this is a method for basically saying "When this observable finalizes, I want to do X (usually perform some clean up) every time" Because you don't know if it will error or complete or just be unsubscribed.

It composes to an observable, because it doesn't cause a subscription. If that makes sense.

```ts

document.on('mousedown').flatMap(() => {
  const div = document.createElement('div');
  document.body.append(div);

  return document.on('mousemove')
    .takeUntil(document.on('mouseup'))
    .finally(() => div.remove())
})
```

domfarolino: > It composes to an observable, because it doesn't cause a subscription. If that makes sense.

This is pretty much what I've been asking about. If it doesn't cause subscription, *and* it passes in its handler function into the operator, then why does it need to compose to anything at all? Why can't it just return undefined. I feel like it's somewhat confusing for it to take in a callback function handler, _and_ also allow subscription handlers: which gets called first? Can anything inside the `finally()` callback function influence what `Observer` callback gets fired on the subscription?

Although in re-asking this, I guess I somewhat figured out the answer to my question: if it returns undefined, then you actually have no way of subscribing to it at all ever, so I guess that doesn't work at all :/

README.md
```diff
@@ -295,6 +295,92 @@ the platform, since it's an easy drop-in wherever you're handling events today.
 The proposed API shape can be found in
 https://wicg.github.io/observable/#core-infrastructure.
 
+```js
+dictionary ObservableEventListenerOptions {
```
domfarolino, line 299: We don't have the IDL in this file anymore. It is in the spec. Please don't add a whole copy of the IDL to here. Just modify the spec file directly.

domfarolino (changes requested):
OK I think I'm sold on this operator, after my latest comment on the open discussion thread. I just need the README IDL / spec stuff to be addressed and we're good.

README.md
domfarolino: I've taken care of this in the latest two commits.

domfarolino (approved):

PR #51, opened 2023-08-29
Add option to pass signal to promise-returning methods.

### LINKED ISSUE #50: Promise APIs should accept configuration options with `signal`
Right now all of the methods that return promises don't accept a means of cancellation.

They should really allow people to abort those promises, which I suppose should reject the returned promise in the same manner that `fetch` does. (With an `AbortError` or a `DOMException`?)

benlesh:
Closes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,10 @@ dictionary Observer {
   AbortSignal signal;
 };
 
+dictionary PromiseOptions {
+  AbortSignal signal;
+};
+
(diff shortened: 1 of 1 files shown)


domfarolino:
The idea is that this signal would be immediately passed to the underlying "subscribe" operation that gets invoked internally in this algorithms, right?

domfarolino (approved):

Issue #26, opened 2023-07-28
Make the Web IDL valid


domenic:
E.g. `subscribe()` is missing a return type, many parameters are missing names.

PR #57, opened 2023-09-06
Use more specific callback types

domfarolino:
This PR satisfies the rest of #26. It uses more specific callback types for the following operators:
 - `map()`
 - `flatMap()`
 - `forEach()`

In particular, it uses the [unsigned long long](https://webidl.spec.whatwg.org/#idl-unsigned-long-long) WebIDL type to represent the "index" of the item being passed into the corresponding callback, since it is the largest _integer_ type that WebIDL is. At first I was worried about that, because that type has a max value of 2<sup>64</sup> - 1, which is strictly larger than ECMAScript's "Number" type, which appears to be used for the same "index" argument in the Iterator helpers (see [`map()`](https://tc39.es/proposal-iterator-helpers/#sec-iteratorprototype.map)) and AsyncIterator helpers (see [`map()`](https://tc39.es/proposal-async-iterator-helpers/#sec-asynciteratorprototype.map)). Specifically, ECMAScript calls 𝔽(counter) for each index that gets incremented, which [maps](https://tc39.es/ecma262/#%F0%9D%94%BD) the index to [a Number](https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type), which is just a double (and smaller than an unsigned long long).

Fortunately, when WebIDL [converts an unsigned long long to an ECMAScript value](https://webidl.spec.whatwg.org/#ref-for-idl-unsigned-long-long%E2%91%A4), it does so in the exact same way that ECMAScript does (by choosing the closest ECMAScript "Number" value to the given value as possible, choosing the even even significand one when necessary).

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,10 @@ interface Subscriber {
 
 callback Predicate = boolean (any value);
 callback Reducer = any (any accumulator, any currentValue)
+callback Mapper = any (any element, unsigned long long index)
+// Differs from `Mapper` only in return type, since this callback is exclusively
+// used to visit each element in a sequence, not transform it.
+callback Visitor = undefined (any element, unsigned long long index)
 
 [Exposed=*]
 interface Observable {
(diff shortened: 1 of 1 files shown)


domenic (approved):
Looks good!

Kind of funny that we don't have that `Visitor` type in specs already, but I guess all the `forEach` methods on the platform are currently automatically generated by `iterable<>` declarations and such.

PR #59, opened 2023-09-07
Describe user needs

### LINKED ISSUE #11: The TAG is going to ask you about "user needs"
See, e.g., this sort of exchange: https://github.com/w3ctag/design-reviews/issues/760#issuecomment-1447787274 

I don't know what the correct answer is here, personally. Maybe just mention that like things such as promises or EventTarget, this enables developers to build websites (particularly ones involving repeated events, which are a common pattern as you can see from ... all web specs ever). And users like websites.

domfarolino:
This closes #11 to the best of my ability, by describing the albeit slightly-tenuous connection this proposal has to direct end user needs.

WDYT @domenic? I tried looking for an AbortController TAG review — since that's an API I think motivated by ergonomics, too, and integrates _with_ this one — and found that there was not one (maybe we weren't doing TAG reviews in the same way back then). The closest I could find was https://github.com/w3ctag/design-reviews/issues/569, which does not mention user needs at all (neither does https://github.com/whatwg/dom/issues/911 it appears).

I'm hoping this will be "good enough".

### DIFF
--- a/README.md
+++ b/README.md
@@ -688,8 +688,8 @@ is a thriving userland Observables ecosystem which results in loads of excessive
 bytes being downloaded every day.
 
 In an attempt to codify the strong userland precedent of the Observable API,
-this proposal would save dozens of custom implementations of it from being
-downloaded every day.
+this proposal would save dozens of custom implementations from being downloaded
+every day.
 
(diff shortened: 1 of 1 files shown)


domenic (approved):
Looks great!

Issue #44, opened 2023-08-11
Methods that accept Observable could accept anything Observable.from does


benlesh:
Given there's an issue to possibly add `Observable.from` to this API #28, and I think that's a good add, it's worth mentioning that in RxJS anywhere we accept an `Observable` we also accept anything that `Observable.from` would accept. 

This is stemming from my [comment about `takeUntil` here](https://github.com/domfarolino/observable/issues/34#issuecomment-1675377433).

Basically anywhere we accept an `Observable` or a function returns an `Observable` that the functionality is processing (`takeUntil` or `flatMap` for example), it's totally fine to also allow anything that `Observable.from` could accept and just do that conversion for the user. It cleans up a lot of code to allow this.

ljharb:
Generally the purpose of a `.from` method is to prevent that overloading - meaning, that users are forced to do the nice, explicit thing, and ensure they have a real Observable before using them as such.

Obviously any method could accept both sets of inputs by wrapping the observable-ish in `Observable.from` but I would not expect builtin methods to do that.

domenic:
Built-in methods accepting anything that can be converted is a common pattern on both the web platform and JS built-ins. +1 to this idea.

ljharb:
Ah, as an argument, that's true - but not as a receiver.

domfarolino:
Closed by https://github.com/domfarolino/observable/pull/60.

PR #60, opened 2023-09-07
`Observable.from()` & accept conversions

### LINKED ISSUE #28: `Iterator.from()` seems to be a thing now
This proposal says

> This appears [in the TC39 proposal's README.md](https://github.com/tc39/proposal-iterator-helpers#fromobject) file but not the spec, so its fate is unclear. leftwards_arrow_with_hook

but https://tc39.es/proposal-iterator-helpers/#sec-iterator.from now exists, as does https://tc39.es/proposal-async-iterator-helpers/#sec-asynciterator.from .

domfarolino:
This PR should close #28 and #44, by providing a static `Observable.from()` method, as well as changing `takeUntil()` — the only web platform method proposed here that takes "an Observable" — to accept the same kinds of objects that `Observable.from()` does.

I had some trouble initially writing this PR. For example: when used as an argument, WebIDL's `Promise<T>` type accepts native Promises as well as all thenables, I believe, which appears to be achieved by delegating to ECMAScript's laxness around thenables. As proposed in this repository, Observables are not part of the ECMAScript language, so they don't get special ES<->WebIDL conversion steps that we might be able to use to accept anything that:
 - Is an `Observable`; or
 - Is a `Promise<T>`; or
 - Has a non-null `Symbol.iterator`; or
 - Has a non-null `Symbol.asyncIterator` protocol

Given that, I was at least hoping we'd be able to supply some kind of nifty conversion algorithm to convert these kinds of types _to_ native Observables, that WebIDL itself could hook into so that any web platform methods that want to accept an Observable can just accept the `Observable` type itself (and the conversion will be done automatically as part of the bindings). Unfortunately I don't think that's possible, nor does it seem possible to express an async iterable type in WebIDL either. For example, https://github.com/whatwg/webidl/issues/874 seems to indicate that the best we can do is accept an `any` and then do the conversion manually by calling ES's `GetIterator(object, async)`, as the Streams Standard does here: https://streams.spec.whatwg.org/#readable-stream-from-iterable.

This isn't horrible, and it doesn't seem like a serious blocker or anything, but if Observables became widely used on the platform we'd be encouraging a bunch of `any` methods that have to delegate to some abstract conversion algorithm that our spec will add. We can always introduce syntax to WebIDL later to make it possible to express a type that is async iterable (something like `async sequence<T>` perhaps); this would allow us to use methods that accept a `typedef (Promise<T> or sequence<T> or async sequence<T>) ObservableInit` — except oh wait, Promises can't be used in unions! So I guess we're stuck with encouraging `any`...

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,8 +325,17 @@ interface Observable {
 
   undefined finally(VoidFunction callback);
 
+  // Constructs a native Observable from `value` if it's any of the following:
+  //   - Observable
+  //   - Promise
+  //   - Iterable
+  //   - AsyncIterable
+  static Observable from(any value);
+
   // Observable-returning operators. See "Operators" section below.
-  Observable takeUntil(Observable notifier);
+  // `takeUntil()` can consume promises, iterables, async iterables, and other
+  // observables.
+  Observable takeUntil(any notifier);
   Observable map(Mapper mapper);
   Observable filter(Predicate predicate);
   Observable take(unsigned long long);
@@ -381,6 +390,21 @@ mechanism](https://dom.spec.whatwg.org/#add-an-event-listener) as
 new event listener whose events are exposed through the Observer handler
 functions and are composable with the various
 [combinators](#operators) available to all Observables.
+
+#### Constructing & converting objects to Observables
+
+Observables can be created by their native constructor, as demonstrated above,
+or by the `Observable.from()` static method. This method constructs a native
+Observable from any of the following objects:
+
+ - `Observable` (in which case it just returns the given object)
+ - `Promise` (or any thenable)
+ - `Iterable` (anything with `Symbol.iterator`)
+ - `AsyncIterable` (anything with `Symbol.asyncIterator`)
+
+Furthermore, any method on the platform that wishes to accept an `Observable`
+can take any of the above objects as well (by accepting a WebIDL `any`), which
+will be converted to a native Observable before being subscribed to.
 
 #### Lazy, synchronous delivery
 
@@ -452,7 +476,10 @@ methods](https://tc39.es/proposal-iterator-helpers/#sec-iteratorprototype) to
  - `some()`
  - `every()`
  - `find()`
- - maybe: `from()`
+
+And the following method statically on the `Iterator` constructor:
+
+ - `from()`
 
 We expect userland libraries to provide more niche operators that integrate with
 the `Observable` API central to this proposal, potentially shipping natively if
--- a/README.md
+++ b/README.md
@@ -413,7 +413,9 @@ ways that we'll finalize in the Observable specification:
     other types.
  2. Require methods and callbacks that work with Observables to specify the type
     `any`, and have the corresponding spec prose immediately invoke a conversion
-    algorithm that the Observable specification will supply.
+    algorithm that the Observable specification will supply. This is similar to
+    what the Streams Standard [does with async iterables
+    today](https://streams.spec.whatwg.org/#rs-from).
 
 The conversation in https://github.com/domfarolino/observable/pull/60 leans
 towards option (1).
--- a/README.md
+++ b/README.md
@@ -404,9 +404,9 @@ Observable from objects that are any of the following, _in this order_:
 
 Furthermore, any method on the platform that wishes to accept an Observable as a
 Web IDL argument, or return one from a callback whose return type is
-`Observable` can do so with any of the above objects as well. This can be
-accomplished in one of two ways that we'll finalize in the Observable
-specification:
+`Observable` can do so with any of the above objects as well, that get
+automatically converted to an Observable. We can accomplish this in one of two
+ways that we'll finalize in the Observable specification:
 
  1. By making the `Observable` type a special Web IDL type that performs this
     ECMAScript Object ➡️ Web IDL conversion automatically, like Web IDL does for
--- a/README.md
+++ b/README.md
@@ -402,9 +402,21 @@ Observable from objects that are any of the following, _in this order_:
  - `Iterable` (anything with `Symbol.iterator`)
  - `Promise` (or any thenable)
 
-Furthermore, any method on the platform that wishes to accept an `Observable`
-can take any of the above objects as well (by accepting a WebIDL `any`), which
-will be converted to a native Observable before being subscribed to.
+Furthermore, any method on the platform that wishes to accept an Observable as a
+Web IDL argument, or return one from a callback whose return type is
+`Observable` can do so with any of the above objects as well. This can be
+accomplished in one of two ways that we'll finalize in the Observable
+specification:
+
+ 1. By making the `Observable` type a special Web IDL type that performs this
+    ECMAScript Object ➡️ Web IDL conversion automatically, like Web IDL does for
+    other types.
+ 2. Require methods and callbacks that work with Observables to specify the type
+    `any`, and have the corresponding spec prose immediately invoke a conversion
+    algorithm that the Observable specification will supply.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -382,6 +391,21 @@ new event listener whose events are exposed through the Observer handler
...
+Observable from any of the following objects:
+
+ - `Observable` (in which case it just returns the given object)
+ - `Promise` (or any thenable)
+ - `Iterable` (anything with `Symbol.iterator`)
+ - `AsyncIterable` (anything with `Symbol.asyncIterator`)
+
+Furthermore, any method on the platform that wishes to accept an `Observable`
+can take any of the above objects as well (by accepting a WebIDL `any`), which
+will be converted to a native Observable before being subscribed to.
```
domfarolino, line 407: This will of course be done in the spec, similarly to what https://streams.spec.whatwg.org/#readable-stream-from-iterable does I imagine.

Also you can imagine passing in an object that satisfies several of the conditions above; e.g., one that's thenable, implements the iterable protocol, _and_ implements the async iterable protocol. In that case, which source which the internal `Obsevable.from()` implementation uses to get the data that it pushes to the created Observable's subscriber is indeed web-observable[^1], so I imagine whatever algorithm that's responsible for doing the conversion must be consistent in which mechanism it prefers (particularly for objects that offer many). That's why in both examples in this PR, I use the same ordering: Observable, Promise, Iterable, AsyncIterable. I think we'd codify this in the spec eventually. Does this at least make sense?

[^1]: That is, whether the `Observable.from()` internals decides to call either `promiseInput.then(next)`, `next(iterableInput[Symbol.iterator]().next())`, or `asyncIterable[Symbol.asyncIterator]().next().then(next)`, if multiple exist.

bakkot:
Note that in principle something can have both a `then` method and a `Symbol.iterator` method, so if you're checking for both, you have to specify which order they're checked.

README.md
domenic: This makes sense. However, I think the order I'd weakly prefer is probably `Observable`, async iterable, iterable, thenable. Reasoning:

- async iterable -> iterable fallback is an established pattern with nice spec support
- Since thenable is a somewhat easier protocol to accidentally fall into, it should be deprioritized in the case of collisions.

README.md
```diff
@@ -382,6 +391,21 @@ new event listener whose events are exposed through the Observer handler
 functions and are composable with the various
 [combinators](#operators) available to all Observables.
 
+#### Constructing & converting objects to Observables
+
+Observables can be created by their native constructor, as demonstrated above,
+or by the `Observable.from()` static method. This method constructs a native
+Observable from any of the following objects:
+
+ - `Observable` (in which case it just returns the given object)
```
domenic, line 400: Interesting. `Array.from(array) !== array`, but it looks like `Iterator.from(iterator) === iterator`. Sticking with the most-recent precedent makes sense, but this is a bit surprising to me.

@bakkot do you know if this asymmetry was discussed and chosen consistently? I feel like I've used `Array.from(array)` as a sort of clone operation in the past.

Maybe the main difference is that iterators and observables are both "immutable" (from the consumer's point of view), whereas arrays are mutable?

domenic:
The big blocker on the Web IDL level is accepting `Promise<T>`s overloaded with anything else. That's explicitly discouraged for most APIs, because promises are special and we want people to be able to pass `x` instead of requiring them to wrap in `Promise.resolve(x)`. However, observables are special too, so it makes sense we'd need an exception here.

My suggestion is that in the spec, you define a conversion operation from a JS value to a Web IDL `Observable` value, using JS spec-type prose. (Like Web IDL does for its conversion algorithms: [example](https://webidl.spec.whatwg.org/#es-sequence).) The first line of any algorithm that takes an "observable-like" should be

1. Set _observableArg_ to the result of [converting to an `Observable`] given _observableArg_.

and then the algorithm can proceed from there assuming _observableArg_ is a Web IDL `Observable`. This quarantines off the JS spec-type prose and value handling into one algorithm, which echoes how Web IDL tries to quarantine off that stuff from the rest of the platform's specs. (Or how the binding layer in implementations quarantines off dealing with the JS engine stuff from the rest of the implementation.)

In the future we can add some sort of `ObservableLike` type to Web IDL which does this step for you. In the meantime, writing the spec this way will set you up for such a nice future.

README.md
bakkot: > do you know if this asymmetry was discussed and chosen consistently?

Yes, though not at length. Iterators are pretty different from arrays - mutability, as you point out, but also you can't _actually_ clone an iterator, since they're stateful. As such we didn't feel the `from` method needed to behave identically (since in fact it can't anyway).

The core of the operation is "give me an iterator from this thing", and there's no reason to make an extra object when the thing you've given is already an iterator with the right prototype. There's no particular benefit, so it would just be wasteful.

annevk:
I suspect we'll want to use that the moment something wants to consume an `Observable`? Which I guess starts with streams? And would we ever want to distinguish between `ObservableLike` and `Observable`? Probably not.

README.md
domfarolino: Done! That sounds good to me. The async iterable -> iterable fallback is news to me, so thanks for chiming in!

domfarolino:
> My suggestion is that in the spec, you define a conversion operation from a JS value to a Web IDL Observable value, using JS spec-type prose.

Yep, that was my plan, I just wasn't sure if that (plus forcing Observable consumers to accept `any`) was a bad practice or felt too hacky, but I think it is fine since at the very least there's precedent for Streams doing it for async iterables. So I think we're "unblocked" here, but I do wonder what the long-term state of affairs might look like. There are two possible follow-ups I can imagine, that seem sort of orthogonal to each other:

 - Making Web IDL allow `Promise<T>` in unions & give Web IDL a syntax for async iterables: this lets Observable consumers express arguments in terms of `ObservableLike` which is a union of a bunch of types. This has some potential downsides I'm not sure of yet:
    - Again, what if a type implements multiple of the characteristics described by the union? Does the union conversion "_prioritize_" conversion order somehow? My understanding is that the spec user is responsible for detecting the active field of the union, like https://fetch.spec.whatwg.org/#concept-bodyinit-extract does, however https://webidl.spec.whatwg.org/#es-union seems to do some preprocessing itself, and eagerly recognizes sequence types, grabbing the `@@iterator`, which I'm not sure we actually want if we're prioritizing async iterables over iterables etc.
 - Making the `Observable` type a special Web IDL type that automatically does the conversion we seek _for us_, but this seems tricky because then it would not only convert things -> Observables for arguments, but would do the same for return values. This is something that I think @annevk mentioned should be avoided in https://matrixlogs.bakkot.com/WHATWG/2023-09-08. I'm not sure how we'd get automatic conversions for Observable arguments, but not Observable return values though.

----

> And would we ever want to distinguish between ObservableLike and Observable? Probably not.

The only thing I can think of is the last point I mentioned above, which is that we'd probably want to allow various things to be passed in as (and converted _to_) an Observable, while preventing code from returning i.e., async iterables in place of an expected Observable return value. At least that's what I thought you were saying on Matrix (and kind of how I interpreted the discussion in https://github.com/domfarolino/observable/issues/44).

annevk:
I think you misunderstood me. Returning `Observable` is reasonable. The problem is with returning "async sequence" and what that might mean. Making `Observable` a special type like `Promise` is probably the best therefore, on reflection. ES-to-IDL would be involved and have "the magic". And IDL-to-ES would be essentially just returning a reference to the JS wrapper object.

This is assuming we'd have many more APIs that want to consume `Observable`. And that generally we'd want to either return a stream type or `Observable` and not some other "async sequence".

domenic (approved):

PR #61, opened 2023-09-12
Add `addTeardown()`. Closes #3. Closes #19.

### LINKED ISSUE #3: The constructor doesn't have a way to register teardown.
This is pretty critical to the design of Observable. There needs to be some sort of way to register setup and teardown during subscription. In RxJS there's a few paths, but the most well-known/understood one is this:

```ts
new Observable(subscriber => {
  // Setup here
  let n = 0
  const id = setInterval(() => subscriber.next(n++), 1000)
 
  return () => {
    // Teardown here
    clearInterval(id)
  }
})
```

However, if we're leaning on DOM APIs like `AbortSignal` we can use that here as well. It's just admittedly more ugly:

```ts
new Observable(subscriber => {
  // Setup here
  let n = 0
  const id = setInterval(() => subscriber.next(n++), 1000)
 
  subscriber.signal.addEventListener('abort', () => {
    // Teardown here
    clearInterval(id)
  })
})
```

Alternatively, it could come in as the second argument, as it's a very commonly accessed piece of functionality:

```ts
new Observable((subscriber, signal) => {
  // Setup here
  let n = 0
  const id = setInterval(() => subscriber.next(n++), 1000)
 
  signal.addEventListener('abort', () => {
    // Teardown here
    clearInterval(id)
  })
})
```

Finally, the `Subscriber` itself could implement event listener, and that might help clean things up:

```ts
new Observable((subscriber) => {
  // Setup here
  let n = 0
  const id = setInterval(() => subscriber.next(n++), 1000)
 
  subscriber.addEventListener('teardown', () => {
    // Teardown here
    clearInterval(id)
  })
})
```

I prefer something like the fourth option, because one of the guarantees observable is supposed to provide is that when you `error()` or `complete()` or `unsubscribe()`, the teardown is called. `"abort"` is maybe a misnomer.

### LINKED ISSUE #19: On using `AbortSignal` instead of `Subscription`
I'm TOTALLY okay with it, and it solves a lot of problems one might have with the "synchronous firehose" issue that comes with dealing with potentially synchronous observables.

However, if we're doing that, you'd want to have the signal passed through the `subscriber` in the constructor in some way, so you can send it along to any inner subscriptions.

It's also worth noting that registering teardown logic with `AbortSignal` is cumbersome and very slow. It might be nice to have a better way to do that.

Finally, if we're going that route, the "teardown return" in the constructor is no longer necessary.

Examples:

```ts
// An observable to stream from a socket
const streamFromSocket = new Observable(subscriber => {
  const socket = new WebSocket('wss://example.com');
  socket.on('open').subscribe({
    next: () => socket.send('start'),
    signal: subscriber.signal,
  });
  socket.on('message').subscribe(subscriber)
})

streamFromSocket.subscribe({
  next: console.log,
  signal: ac.signal,
})


// An observable to stream from someone's custom type
const customData = new Observable(subscriber => {
  const customDataSource = new CustomDataSource();
  customDataSource.onData = e => subscriber.next(e);

  // This is REALLY unergonomic and doesn't account for
  // subscribers that are already "closed".
  subscriber.signal.addEventListener('abort', () => {
    customDataSource.destroy()
  })
})
```

We may wish to add a custom method to `Subscriber` that does that abort registration for users:

```ts
// An observable to stream from someone's custom type
const customData = new Observable(subscriber => {
  const customDataSource = new CustomDataSource();
  customDataSource.onData = e => subscriber.next(e);

  subscriber.addTeardown(() => {
    customDataSource.destroy()
  })
})
```

domfarolino:
This PR clarifies the `addTeardown()` API, which is just syntactic sugar over responding accordingly to `Subscriber#signal`. I'm honestly not 100% sure if we need it, but it does seem like a nice first-class convenience API to get around the pesky is-this-signal-already-aborted check that would have to be made by ever subscriber with a clean-up function, at least until `AbortSignal` gets an API [like `whenAborted()`](https://github.com/domfarolino/observable/issues/19#issuecomment-1700309108).

My only concern here is that it might be confusing if we expose both `Subscriber#signal` _and_ `Subscriber#addTeardown()`, when you really only need one or the other. If we're going to have this first-class API, I kind of lean towards **not** exposing `Subscriber#signal` anymore, and pushing people to use `addTeardown()` exclusively, but what do you think @benlesh?

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,7 @@ If the subscriber has already been aborted (i.e., `subscriber.signal.aborted` is
  - From `complete()`, after the subscriber's complete handler (if any) is
    invoked
  - From `error()`, after the subscriber's error handler (if any) is invoked
+ - The signal passed to the subscription is aborted by the user.
 
 
 ### Operators


benlesh (approved):

README.md
```diff
@@ -470,6 +470,7 @@ If the subscriber has already been aborted (i.e., `subscriber.signal.aborted` is
  - From `complete()`, after the subscriber's complete handler (if any) is
    invoked
  - From `error()`, after the subscriber's error handler (if any) is invoked
+ - The signal passed to the subscription is aborted by the user.
```
domfarolino, line 473: Technically I don't think this was needed because of https://github.com/domfarolino/observable/pull/61/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R466-R467 (i.e., we're already in the "Otherwise" branch here) but I guess it can't hurt?

PR #63, opened 2023-09-14
format markdown file with prettier

benlesh:
Formatting the code in the markdown file with Prettier.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+    "useTabs": true,
+    "singleQuote": true,
+    "trailingComma": "all"
+}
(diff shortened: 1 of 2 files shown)


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
domfarolino, line 1: Does this file just apply to the entire directory? At some point we'll have an actual spec that it shouldn't apply to. Is there a way to specify that? Sorry, I don't know a lot about prettier.

domfarolino: The spec will be a `.bs` file, so that's good to know.

domfarolino (approved):
I think I still disagree with a lot of the formatting quirks here, but I guess don't feel _that_ strongly about them.

Issue #76, opened 2023-10-04
Do we need the `subscriber.closed` attribute?


domfarolino:
In a [review of some Observable constructor web platform tests](https://github.com/web-platform-tests/wpt/pull/42219#discussion_r1341427510), the `Subscriber#closed` attribute was brought up. I hadn't heard of this before, and while I can imagine this could be useful, a concrete practical use-case doesn't really jump out at me, so I'm filing this to spur some discussion about why this might be useful / if it is really needed.

Please see all of the discussion starting at https://github.com/web-platform-tests/wpt/pull/42219#discussion_r1341427510 and https://github.com/web-platform-tests/wpt/pull/42219#discussion_r1361243283.

domfarolino:
https://github.com/web-platform-tests/wpt/pull/42219#discussion_r1361243283 has me just a little confused. It sounds like the AbortSignal passed into `subscribe()` is expected to be different than the AbortSignal that the Observable can pluck off of `Subscriber#signal`. That seems fine, but the purpose of this is to ensure that `subscriber.signal.aborted` is `true` whenever the observable _itself_ completes the subscription. Assuming `subscriber.signal.aborted` is _also_ true when the subscriber aborts its subscription, then it seems that `subscriber.signal.aborted` will **always** equal `subscriber.closed`, in which case I wonder if we have a strong reason to introduce this new API/state.

@benlesh am I right in thinking that `subscriber.signal.aborted === subscriber.closed`?

domfarolino:
I've cleaned up the code example in https://github.com/web-platform-tests/wpt/pull/42219#discussion_r1341838611 that is trying to demonstrate the need for `Subscriber#complete`:

```js
let firstTime = true;
const source = new Observable(subscriber => {  
  if (firstTime) {
    firstTime = false;
    source.subscribe({
      complete: () => {
        if (!subscriber.signal.aborted) {
          subscriber.complete();
        }
      },
      signal: subscriber.signal
    })
  } else {
    subscriber.complete();
  }
})

const ac = new AbortController();
source.subscribe({
  complete: () => console.log('complete!'),
  signal: ac.signal,
});
```
In Chromium with the Observables flag enabled, it seems to run fine / without issue. So I'm having a hard time figuring out why we need `.closed`. Sorry if I'm just being dense!

domfarolino:
[...]

Issue #22, opened 2023-06-06
Is there any concern about the difference in constructor semantics from `Promise`?


tbondwilkinson:
To preface, I understand that RxJS and other Observable implementations are widely used and very popular, and that a lot of people have been very successfully using Observables for  very long time in many applications.

But there's a difference between using a library and using a web platform primitive. And I do worry a little that people will expect the `Observable` constructor to behave similarly to the `Promise` constructor, in particular being executed synchronously and only once for every `.then()` callback. Observables don't really work that way, as the callback parameter is called when `subscribe` is called, an unlimited number of times. This is the much discussed cold vs hot distinction.

So for example, maybe it's really easy to write:

```
const observable = new Observable((subscriber) => {
  subsriber.next(someReallyExpensiveComputation());
});
```

And people comfortable with promises will think that's only going to be executed once and then each subscriber gets the expensive value, when in reality the expensive computation happens once for every subscriber.

This is less of a concern for Observables that the platform creates for something like event listeners, where the browser is in charge of the observable instantiation and the rest of the API is just really offering a better listener pattern.

benlesh:
>  I do worry a little that people will expect the Observable constructor to behave similarly to the Promise constructor, in particular being executed synchronously and only once for every .then() callback.

That cat is sort of already out the the bag for more than a decade now. It's a non-issue in terms of broad understanding of the type. I know that's dismissive, so I'll try to elaborate.

However, to play devil's advocate, there are other APIs that don't quite line up with what the `Promise` constructor does despite having a similar "look", for example `MutationObserver` or `ResizeObserver`:

```ts
const resizeObserver = new ResizeObserver((entries) => {
  // do stuff
});

resizeObserver.observe(element);

resizeObserver.disconnect();

const mutationObserver = new MutationObserver((entries) => {
  // do stuff
});

mutationObserver .observe(element);

resizeObserver.disconnect();
```

And honestly, those two APIs would have been _so much better off_ as observables, because as they're currently implemented, while you can `unobserve` different elements, if you `observe` two different elements because you want to do something similar with each, you end up needing to add conditionals in the callback to handle them in whatever slightly different ways you might need to. As opposed to something like `new MutationObservable(element).subscribe(callback)` or the like.

---

In terms of unicast/multicast, which is this issue: 

```ts
const expensiveThingObservable = new Observable((subscriber) => {
  subscriber.next(someReallyExpensiveComputation());
});
```

This is already something people face with ordinary functions:

```ts
[...]

PR #80, opened 2023-11-10
Introduce `Subscriber#active` boolean

domfarolino:
This closes https://github.com/WICG/observable/issues/76. This PR introduces the `subscriber.active` boolean attribute that can be used from within an Observable to determine whether a subscription is still active. I am still personally a tiny but unsure whether this is really necessary, so I'd like to use this space to describe the semantics and how it differs from other APIs, so we can come to a conclusion on this, since @benlesh feels much more strongly about this API.

On the surface it seems that `subscriber.active === !subscriber.signal.aborted`[^1], however this is not quite true. The lifecycle of an Observable is the following:

1. Active subscription
2. `subscriber.complete()/error()` is called from script
3. The subscription is no longer "active", meaning the bodies of those methods cannot re-entrantly invoke themselves or _any_ method on the `Observer` dictionary for that matter.
4. The actual corresponding `Observer#complete()/error()` body is run
5. The `Subscriber#signal` is aborted (via an internal `AbortController`)
6. `subscriber.signal.aborted` [is true](https://dom.spec.whatwg.org/#ref-for-abortsignal-abort-reason%E2%91%A0%E2%91%A2)
7. All teardown callbacks (added by an API [that probably looks like](https://github.com/WICG/observable/issues/22#issuecomment-1805840402) `subscriber.addTeardown()`) will run, as part of the abort algorithms associated with `subscriber.signal`

The `subscriber.active` attribute intends to expose precisely when a subscription no longer becomes active (just before the `Observer#complete()/error()` body is run), which is _before_ the `subscriber.signal` is finally aborted. The only observable consequence I can think of for this is _inside_ the `complete()`/`error()` methods, where `subscriber.active` is already false but `subscriber.signal.aborted` is not yet true (since teardowns haven't run yet).

I think that `subscriber.active` is a more sensible and ergonomic way to express the concept of an "inactive subscription", however I only hesitate to add it because I don't love the idea of having two _different_ ways of getting this information with slightly different semantics. I'm curious if anyone would rely on `subscriber.active` in a way that `!subscriber.signal.aborted` couldn't be used for (thus causing broken code).

----

Here's an I can think of where `subscriber.signal.aborted` is insufficient for stopping an infinite loop:

```js
let stopPushingValues = true;

const observable = new Observable(subscriber => {
  // The subscriber's `complete()` handler fires an event that
  // forces this observable to synchronously push a firehose of
  // values to the subscriber.
  document.addEventListener('stuff', e => {
    while (!subscriber.signal.aborted) {
      if (stopPushingValues === false) {
        subscriber.next(1);
      } else {
        // In this case, the subscriber has signaled something
        // to us that it wants us to complete the subscription.
        subscriber.complete();
      }
    }
  });
});

observable.subscribe({
  next: v => {
    stopPushingValues = true;
  },
  complete: () => {
    document.dispatchEvent(new Event('stuff'));
  }
});

document.dispatchEvent(new Event('stuff'));

// The sequence of events is this:
//   1. Random code fires a 'stuff' event at document
//   2. This causes the Observable to start pushing values synchronously
//      to the `next()` handler. Eventually the `next()` handler decides that
//      doesn't want anymore values, so it signals to the Observable that it
//      wants the Observable to call `complete()`[^1] to end the subscription.
//   3. Observable detects `stopPushingValues` (i.e., that the subscriber wants
//      the Observable to stop), and calls `Observer#complete()`.
//   4. Observer#complete() triggers another 'stuff' event, which re-enters the
//      event-pushing code in the subscriber callback. This starts another
//      synchronous loop that tries to push values *again* to the subscriber. But
//      this time no values can be pushed, because once `complete()` is entered ever,
//      it becomes impossible to invoke `Observer#next()/complete()/error()` methods.
//      But the `while` loop never stops!
//
//      So here's how you could get yourself into this buggy situation. After you call
//      `complete()`, it's true that `subscriber.signal.aborted` will be true. As a
//      developer, you might rely on this, by using `while (!subscriber.signal.aborted)`
//      as a condition in the while loop, so that the while loop breaks *after* you call
//      `complete()`. However there's a very subtle timing thing here that you might not
//      realize. `subscriber.signal.aborted` is only true *after* the call to `complete()`
//      is done. So if `complete()` itself re-enters the while loop, it will never terminate
//      because `subscriber.signal.aborted` is never `true` *from within* the call to
//      `complete()`.

// [^1]: Note that this is the contrived part of this example. Normally when a
// subscriber wants to be "done", it just `abort()`s the signal that was passed into
// `subscribe()`, to unsubscribe. But that doesn't cause the issue in this example,
// so we had to craft a more contrived way to trigger it.
```

How important is it to protect against this kind of (highly contrived!) scenario? Or are there way less contrived scenarios where `subscriber.signal.aborted` is a footgun, and where `subscriber.active` would be critical.

[^1]: Remember of course that `subscriber.signal` is aborted _absolutely whenever_ a subscription is ended, not just when the subscriber unsubscribes / aborts the `signal` that it passed into `subscribe()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,11 @@ interface Subscriber {
   undefined error(any error);
   undefined addTeardown(VoidFunction teardown);
 
+  // True after the Subscriber is created, up until either
+  // `complete()`/`error()` are invoked, or the subscriber unsubscribes. Inside
+  // `complete()`/`error()`, this attribute is true.
+  readonly attribute boolean active;
+
   readonly attribute AbortSignal signal;
 };
 


benlesh (approved):

benlesh:
Yeah. This API in RxJS is called `closed`, and used fairly often, but `active` is a MUCH better name for it, as most checks are like `if (subscriber.active) { /* do things */ }`, and because it better describes what you're checking. You're checking to see if the subscriber is still "active" before you continue doing work.

domfarolino:
Replaced with https://github.com/WICG/observable/pull/82 since this was merged by mistake.

PR #86, opened 2023-11-21
Adds explaination of `flatMap` semantics

benlesh:
Basically this is describing RxJS's `concatMap`. I tried to add it as detailed as possible because "It works like `Iterator.prototype.flatMap` is only partially true, since there's the element of asynchrony involved here. If all sources and inner sources are synchronous, though, it is exactly like `Iterator.prototype.flatMap`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,6 +507,50 @@ We propose the following operators in addition to the `Observable` interface:
 - `finally()`
   - Like `Promise.finally()`, it takes a callback which gets fired after the
     observable completes in any way (`complete()`/`error()`)
+- `flatMap()`
+
+  - Similar to `Iterator.prototype.flatMap`, however, because the types are different,
+    there are some semantics to note.
+
+    Given the following example:
+
+```ts
+const result = source.flatMap((value) => getNextInnerObservable(value));
+```
+
+1. When you subscribe to `result`, it subscribes to `source` immediately.
+2. Let there be a queue of values that is empty.
+3. Let there be an `innerSignal` that is either `undefined` or an `AbortSignal`
+4. Let there be an `isSourceComplete` that is `false`.
+5. When the `source` emits a value:
+   a. If `innerSignal` is `undefined`
+   1. Call the mapping function with the the value.
+   2. Then pass the return value of the mapping function to `Observable.from()` to convert it to
+      an "inner observable" if it's not already.
+   3. Then create an `AbortSignal` that follows the subscriber's
+      signal, and set `innerSignal`.
+   4. pass the `innerSignal` to the subscribe for the inner observable.
+      emit all values from the inner observable to the `result` observer.
+   5. If the inner observable completes, check to see if there are any queued values.
+      a. If there is, take the first one from the queue, and move to 5.a.1.
+      b. If there's not
+      1. If `isSourceComplete` is `true`
+         a. Complete `result`.
+      2. If `isSourceComplete` is `false`
+         a. Move to 5 and wait.
+   6. If the inner observable errors:
+      a. Forward the error to `result`.
+      b. If `innerSignal` is `AbortSignal`
+   7. Add the value to the queue and wait.
+6. If the `source` completes:
+   a. If `innerSignal` is `undefined`
+   1. Complete `result`.
+      a. If `innerSignal` is `AbortSignal`
+   1. set `isSourceComplete` to `true`.
+7. If the `source` errors:
+   a. Forward the error to `result`.
+8. If the user aborts the signal passed to the subscription of `result`
+   a. Abort any `innerSignal` that exists, and terminate subscription.
 
 Versions of the above are often present in userland implementations of
 observables as they are useful for observable-specific reasons, but in addition
--- a/README.md
+++ b/README.md
@@ -512,46 +512,6 @@ We propose the following operators in addition to the `Observable` interface:
   - Similar to `Iterator.prototype.flatMap`, however, because the types are different,
     there are some semantics to note.
 
-    Given the following example:
-
-```ts
-const result = source.flatMap((value) => getNextInnerObservable(value));
-```
-
-- When you subscribe to `result`, it subscribes to `source` immediately.
-- Let there be a `queue` of values that is empty.
-- Let there be an `innerSignal` that is either `undefined` or an `AbortSignal`
-- Let there be an `isSourceComplete` that is `false`.
-- When the `source` emits a value:
-  - If `innerSignal` is `undefined`
-    - Call the mapping function with the the value.
-    - Then pass the return value of the mapping function to `Observable.from()` to convert it to
-      "inner observable" if it's not already.
-    - Then create an `AbortSignal` that follows the subscriber's and set `innerSignal`.
-    - pass the `innerSignal` to the subscribe for the inner observable.
-    - Forward all values emitted by the inner observable to the `result` observer.
-    - If the inner observable completes
-      - If there are values in the `queue`
-        - take the first one from the `queue` and return the the mapping step.
-      - If the `queue` is empty
-        - If `isSourceComplete` is `true`
-          - Complete `result`.
-        - If `isSourceComplete` is `false`
-          - Wait
-    - If the inner observable errors
-      - Forward the error to `result`.
-  - If `innerSignal` is `AbortSignal`
-    - Add the value to the `queue` and wait.
-- If the `source` completes:
-  - If `innerSignal` is `undefined`
-    - Complete `result`.
-  - If `innerSignal` is `AbortSignal`
-    - set `isSourceComplete` to `true`.
-- If the `source` errors:
-  - Forward the error to `result`.
-- If the user aborts the signal passed to the subscription of `result`
-  - Abort any `innerSignal` that exists, and terminate subscription.
-
 Versions of the above are often present in userland implementations of
 observables as they are useful for observable-specific reasons, but in addition
 to these we offer a set of common operators that follow existing platform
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -507,6 +507,10 @@ We propose the following operators in addition to the `Observable` interface:
 - `finally()`
   - Like `Promise.finally()`, it takes a callback which gets fired after the
     observable completes in any way (`complete()`/`error()`)
+- `flatMap()`
```
domfarolino, line 510: I don't think this should appear both in this "exceptional" list, _and_ in the list below about iterator helpers. Seems like it should just exist in the list of iterator helpers, with a note that the semantics are a little different (as you provided below).

README.md
```diff
@@ -554,6 +554,55 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+
+Given the following example:
+
+```ts
+const result = source.flatMap((value) => getNextInnerObservable(value));
+```
+
+- When you subscribe to `result`, it subscribes to `source` immediately.
+- Let there be a `queue` of values that is empty.
+- Let there be an `innerSignal` that is either `undefined` or an `AbortSignal`
```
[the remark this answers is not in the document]
benlesh: That's correct. I'll add something.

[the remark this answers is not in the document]
benlesh: Actually, I do explain that below: `Pass the `innerSignal` to the subscribe for the inner observable.`

README.md
```diff
@@ -554,6 +554,55 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+- Let there be an `isSourceComplete` that is `false`.
+- When the `source` emits a value:
+  - If `innerSignal` is `undefined`
+    - Call the mapping function with the the value.
```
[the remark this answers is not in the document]
benlesh: Adding detail.

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
 means event handlers that call `e.preventDefault()` will run too late. See the
 [Concerns](#concerns) section which goes into more detail.
 
+### flatMap semantics
+
+`flatMap` generally behaves like `Iterator.prototype.flatMap`, however, since it's push-based,
+there can be a temporal element. Given that, it behaves much like RxJS's `concatMap`, which is
+one of the most useful operators from the library.
```
domfarolino, line 561: > which is one of the most useful operators from the library.

I think we can remove this.

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+
+At a high-level, it subscribes to the source, and then maps to, and emits values from "inner
+observables", one at a time, ensuring they're subscribed to in sequence.
+
+Given the following example:
+
+```ts
+const result = source.flatMap((value, index) =>
+	getNextInnerObservable(value, index),
```
domfarolino, line 570: What's the trailing comma here?

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+```
+
+- `flatMap` is a method on `Observable` that is called with a `mapping function`, which takes a
+  value from the source observable, and a zero-based counter (or "index") of that value, and
+  returns a value that can be converted to an observable with `Observable.from`. `flatMap` returns
+  an obeservable we'll call `result`.
+- When you subscribe to `result`, it subscribes to `source` immediately.
+- Let there be a `queue` of values that is empty.
+- Let there be an integer `current index` that is `0`.
+- Let there be an `innerSignal` that is either `undefined` or an `AbortSignal`.
```
domfarolino, line 581: initially....? Where does this come from? It comes from the subscribe() method being called on the Observable returned by flatMap(), right? Maybe we should mention that. After reading the rest I'm actually thinking that `innerSignal` is not the one passed into subscribe. So... I'm not sure :)

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+- Let there be an `isSourceComplete` that is `false`.
```
domfarolino, line 582:
```suggestion
- Let there be a boolean `isSourceComplete` that is `false`.
```

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+- When the `source` emits a `value`:
+  - If `innerSignal` is `undefined`
+    - Begin **"mapping step"**:
+      - Copy the `current index` into an `index` variable.
+      - Increment the `current index`.
+      - Call the `mapping function` with the the `value` and the `index`.
```
domfarolino, line 588:
```suggestion
      - Call the `mapping function` with `value` and `index`.
```

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+Given the following example:
+
+```ts
+const result = source.flatMap((value, index) =>
+	getNextInnerObservable(value, index),
+);
+```
+
+- `flatMap` is a method on `Observable` that is called with a `mapping function`, which takes a
+  value from the source observable, and a zero-based counter (or "index") of that value, and
```
domfarolino, line 575:
```suggestion
  value from the source observable, and a zero-based counter (or `index`) of that value, and
```

Just to make this consistent with `mapping function`

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+- Let there be an integer `current index` that is `0`.
+- Let there be an `innerSignal` that is either `undefined` or an `AbortSignal`.
+- Let there be an `isSourceComplete` that is `false`.
+- When the `source` emits a `value`:
+  - If `innerSignal` is `undefined`
+    - Begin **"mapping step"**:
+      - Copy the `current index` into an `index` variable.
+      - Increment the `current index`.
+      - Call the `mapping function` with the the `value` and the `index`.
+      - Then pass the return value of the mapping function to `Observable.from()` to convert it to
```
domfarolino, line 589:
```suggestion
      - Then pass the return value of the `mapping function` to `Observable.from()` to convert it to
```

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+        "inner observable" if it's not already.
+      - Then create an `AbortSignal` that is dependent on the subscriber's and set `innerSignal`.
```
domfarolino, line 591:
```suggestion
      - Then create an `AbortSignal` that is [dependent](https://dom.spec.whatwg.org/#create-a-dependent-abort-signal) on the subscriber's and set `innerSignal`.
```

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+      - Pass the `innerSignal` to the subscribe for the inner observable.
```
domfarolino, line 592:
```suggestion
      - Subscribe to the inner observable, passing `innerSignal`
```

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+      - Forward all values emitted by the inner observable to the `result` observer.
+      - If the inner observable completes
```
domfarolino, line 594:
```suggestion
      - If the inner observable completes;
```

assuming you meant to coalesce this and the following condition.

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+      - If there are values in the `queue`
+        - Take the first one from the `queue` and return to the **"mapping step"**.
+      - If the `queue` is empty
+        - If `isSourceComplete` is `true`
+        - Complete `result`.
+        - If `isSourceComplete` is `false`
+        - Wait
+      - If the inner observable errors
+        - Forward the error to `result`.
+  - Otherwise, if `innerSignal` is `AbortSignal`
```
domfarolino, line 604:
```suggestion
  - Otherwise, if `innerSignal` is an `AbortSignal`
```

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+        "inner observable" if it's not already.
+      - Then create an `AbortSignal` that is dependent on the subscriber's and set `innerSignal`.
+      - Pass the `innerSignal` to the subscribe for the inner observable.
+      - Forward all values emitted by the inner observable to the `result` observer.
+      - If the inner observable completes
+      - If there are values in the `queue`
+        - Take the first one from the `queue` and return to the **"mapping step"**.
+      - If the `queue` is empty
+        - If `isSourceComplete` is `true`
+        - Complete `result`.
```
domfarolino, line 599: Sometimes you have a period and sometimes you don't. Can you edit this whole algorithm to be consistent one way or another?

README.md
```diff
@@ -554,6 +554,65 @@ Promises whose scheduling differs from that of Observables, which sometimes
...
+        - If `isSourceComplete` is `false`
+        - Wait
```
domfarolino, line 601: Hmm, what is this supposed to do? Certainly we're not blocking the main thread....

benlesh: No... Just "wait for another event"... do nothing, I guess.

README.md
benlesh: Maybe I can.
Maybe I can't

README.md
benlesh: Probably aggressive formatting rules. I'll have a look.

README.md
benlesh: It's a bit of state for the internal algorithm. If `innerSignal` is `undefined`, there's no current inner subscription. If `innerSignal` is an `AbortSignal`, we have an inner subscription, and we need to wait for it to complete before we start another. So we need to buffer the value.

Basically:

```ts
let innerSignal: AbortSignal | undefined;
```

domfarolino:
A few outstanding comments from a while ago still.. I'll wait for those to get resolved.

domfarolino:
Since this operator has already made it into the spec https://wicg.github.io/observable/#dom-observable-flatmap, I think this may not be necessary, and seems to have no activity these days anyways. So I'll close it.

Issue #64, opened 2023-09-15
I think `catch` may be a requirement


benlesh:
We currently aren't providing a way to gracefully handle an error. We do have error handling in the subscription/observer itself, but no way to have the subscription "recover" from a fatal error.

Since subscriptions have ended the moment an error is emitted from an observable, the `catch` method behaves a lot like `catch` on promise, meaning that it returns an observable that will switch to whatever observable you map to in its handler in the event of an error.

For example:

```ts
// Given this observable that will always error:
const source = button1.on('click').map((e, i) => {
  if (i === 2) {
    throw new Error('boom!');
  }
  return e;
});


// If we catch the error and return a different observable,
// the new observable takes over the subscription.
source.catch((err) => {
  console.error(err);
  // let another button take over
  return button2.on('click');
})
.subscribe({
  next: console.log,
  error: () => { /* never called */ }
});

// If we catch the error and return the same observable, you can
// resume or "retry" that observable
source.catch((err) => {
  console.log(err);
  return source;
})
.subscribe({
  next: console.log,
  error: () => { /* never called */ }
});

// If we catch the error and throw it again, or throw any other error
// The resulting observable still errors
source.catch((err) => {
  throw new Error('Different error!');
})
.subscribe({
  next: console.log,
  error: console.error
});
```

PR #87, opened 2023-11-22
Adds the catch operator to the README

benlesh:
Makes progress on https://github.com/WICG/observable/issues/64. Simply adds the `catch` operator to the `README`. This is one of the most widely used operators. It's particularly useful in preventing inner subscriptions from terminating outer subscriptions in cases like long-polling or events that kick off other event streams.

For example, let's say someone creates an observable from a web socket:

```ts
function socket(url) {
  return new Observable(subscriber => {
    const ws = new WebSocket(url);
    
    subscriber.addTeardown(() => {
      if (ws.readyState <= WebSocket.OPEN) {
        ws.close();
      }
    });
    
    ws.onopen = (e) => {
      subscriber.next(e);
    };
    
    ws.onmessage = (e) => {
      subscriber.next(e);
    };
    
    ws.onerror = (e) => {
      subscriber.error(e);
    };
    
    ws.onclose = (e) => {
      if (e.wasClean) {
        subscriber.complete();
      } else {
        subscriber.error(e);
      }
    };
  })
}

const socketStream = socket('wss://someendpoint')

socketStream
  .do(e => {
    if (e.type === 'open') {
      const socket = e.target;
      socket.send('start');
    }
  })
  .filter(e => e.type === 'message')
  .map(e => JSON.parse(e.data))
  .subscribe(console.log);
```

The above stream will terminate the socket under the following error conditions:

1. The socket closes "dirty", often due to losing network.
2. The socket experiences another error
3. The message data fails `JSON.parse`.

Ideally, in the first two cases, we can reconnect the socket. To do this, we can use `catch`:

```ts
// A utility for timeout
function timeout(ms) {
  return new Observable(subscriber => {
    const id = setTimeout(() => {
      subscriber.next();
      subscrber.complete();
    }, ms);
    subscriber.addTeardown(() => clearTimeout(id));
  });
}

// This is now a socket that will retry if it fails.
const socketStream = socket('wss://someendpoint')
  .catch(err => {
    // log the error
    console.error(err);
    
    if (err.type === 'close' && !err.wasClean) {
      if (!navigator.onLine) {
        // Wait until we're online and reconnect
        return window.on('online')
          .take(1)
          .flatMap(() => socketStream)
      }
    }
    // Wait for 1 second and try again
    return timeout(1000)
      .flatMap(() => socketStream);
  });

socketStream
  .do(e => {
    if (e.type === 'open') {
      const socket = e.target;
      socket.send('start');
    }
  })
  .filter(e => e.type === 'message')
  .map(e => {
     // These sorts of errors are generally going to be
     // resolved the traditional way.
     try {
       return JSON.parse(e.data)
     } catch (err) {
       console.error(err);
       return { parseError: err }
     }
  })
  .subscribe(console.log);
```


domfarolino:
This seems to add back all of the IDL to the README file. Maybe this means to instead just add a couple lines to the README, and a few more to the spec?

README.md
```diff
@@ -416,6 +416,10 @@ If the subscriber has already been aborted (i.e., `subscriber.signal.aborted` is
 
 We propose the following operators in addition to the `Observable` interface:
 
+- `catch()`
+  - Like `Promise.catch()`, it takes a callback which gets fired after the source
```
domfarolino, line 420:
```suggestion
  - Like `Promise#catch()`, it takes a callback which gets fired after the source
```

Super nit: just since this is not a static method I want to use the weird `#` syntax to indicate that you can't actually open up DevTools and just call `Promise.catch()` like the example here makes it look like.

domfarolino (approved):

PR #97, opened 2024-01-05
Spec the `toArray()` operator

domfarolino:
This PR finishes spec'ing the `toArray()` operator, as a warm-up for the rest.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/97.html" title="Last updated on Jan 11, 2024, 3:13 PM UTC (48c33e5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/97/457ba36...48c33e5.html" title="Last updated on Jan 11, 2024, 3:13 PM UTC (48c33e5)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -605,9 +605,6 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 <div algorithm>
   The <dfn for=Observable method><code>toArray(|options|)</code></dfn> method steps are:
 
-    1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
-       Document=] is not [=Document/fully active=], then return.
-
     1. Let |p| [=a new promise=].
 
     1. If |options|'s {{SubscribeOptions/signal}} is not null:
--- a/spec.bs
+++ b/spec.bs
@@ -34,6 +34,8 @@ urlPrefix: https://dom.spec.whatwg.org; spec: DOM
       text: passive; url: event-listener-passive
       text: once; url: event-listener-once
       text: signal; url: event-listener-signal
+    for: AbortSignal
+      text: dependent signals; url: abortsignal-dependent-signals
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


domfarolino:
@benlesh Would you be able to take a look at this and confirm that the semantics look right to you?

domfarolino:
I'm trying to see why PR preview isn't working on this PR. Closing and re-opening to see if that triggers it.

spec.bs
```diff
@@ -605,25 +605,31 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 <div algorithm>
   The <dfn for=Observable method><code>toArray(|options|)</code></dfn> method steps are:
 
+    1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
+       Document=] is not [=Document/fully active=], then return.
```
domenic, line 609: This isn't valid; you need to return a promise.

Also, I don't think you should have this kind of condition? There's no dependency on the document being fully active here.

domfarolino: > you need to return a promise.

Oops, yes!

> Also, I don't think you should have this kind of condition? There's no dependency on the document being fully active in the following steps.

Well, I thought we shouldn't be explicitly trying to resolve / reject promises if the document isn't active, right? And that's what the following steps are set up to do. Or did we land on resolving/rejecting being a no-op, so we can just do it without gating on this kind of condition.

domenic: I generally land on it being a no-op. Like, nothing prevents web developers from doing

```js
const iframeWindow = iframe.contentWindow;
iframe.remove();
const { promise, resolve } = iframeWindow.Promise.withResolvers();
resolve("foo");
```

so especially for "foundational" things like streams or observables, I feel like we should not add extra guards.

domfarolino: OK sounds good, I've removed the condition. Thanks.

PR #101, opened 2024-01-10
rename "Observer" to "SubscriptionObserver"

### LINKED ISSUE #99: Are IDL names like `Observer` too ambiguous?
Looking through the spec here, I see `Observer` is an IDL type but that might be a little ambiguous, as there are lots of "Observer" like things that exist across various specs, for example `ResizeObserver`, `MutationObserver` and so on.

This has practical ramifications, as [Webkit already has an Observer class](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/Observer.h), and [Firefox also has an Observer class](https://searchfox.org/mozilla-central/source/xpcom/ds/Observer.h) that are for different things. It would be useful to disambiguate against those, and it might be easier to simply rename this specifications `Observer` type.

Might I suggest a few alternative names:

 - `SubscriptionObserver`
 - `SubscriberLike` (my preference)
 - `ObservableSubscriber`
 - `SubscribeCallbacks`
 - `ObserverCallbacks`

keithamus:
Closes #99

***
<a href="https://pr-preview.s3.amazonaws.com/keithamus/observable/pull/101.html" title="Last updated on Jan 17, 2024, 9:22 PM UTC (46743cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/101/457ba36...keithamus:46743cb.html" title="Last updated on Jan 17, 2024, 9:22 PM UTC (46743cb)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -418,10 +418,11 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
 
 <div class=note>
   <p>The [=internal observer=] [=struct=] is used to mirror the {{SubscriptionObserver/next}},
-  {{SubscriptionObserver/error}}, and {{SubscriptionObserver/complete}} [=callback functions=]. For any {{Observable}} that
-  is subscribed by JavaScript via the {{Observable/subscribe()}} method, these algorithm "steps"
-  will just be a wrapper around [=invoking=] the corresponding {{SubscriptionObserver/next}},
-  {{SubscriptionObserver/error}}, and {{SubscriptionObserver/complete}} [=callback functions=] provided by script.</p>
+  {{SubscriptionObserver/error}}, and {{SubscriptionObserver/complete}} [=callback functions=]. For
+  any {{Observable}} that is subscribed by JavaScript via the {{Observable/subscribe()}} method,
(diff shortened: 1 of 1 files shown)


domfarolino:
The IPR bot is upset, so you could you follow the instructions on https://labs.w3.org/repo-manager/pr/id/WICG/observable/101, and join the WICG so the bot will be happy with this and your future contributions?

spec.bs
```diff
@@ -309,13 +309,13 @@ console.assert(producedValues.length === 0);
...
 
-dictionary Observer {
+dictionary SubscriptionObserver {
   ObserverCallback next;
   ObserverCallback error;
   VoidFunction complete;
 };
 
-typedef (ObserverCallback or Observer) ObserverUnion;
+typedef (ObserverCallback or SubscriptionObserver) ObserverUnion;
```
domfarolino, line 318: I wonder if we should rename `ObserverCallback` to `SubscriptionObserver` callback, in the spirit of specificity and to keep these consistent?

keithamus: I've done that, let me know what you think.

domfarolino (approved):

keithamus:
@domfarolino all green now 🙌

Issue #69, opened 2023-09-21
Change Promise returning methods to return Observable (except forEach)


benlesh:
### Promises bearing events have problems

Because promises force scheduling, any API that returns a promise runs the risk of confusing users because:

1. It causes subscription, where other methods do not.
2. The promise fires asynchronously, meaning it's too late to `e.preventDefault()`.

Basically, we should have everything accept `forEach` return an `Observable`.

### Semantics of forEach

`forEach` would still synchronously execute its callback, but the completion being a promise would mean completion is scheduled. This is okay, because completion doesn't bear a value.

```ts
const et = new EventTarget();
  
const testOver = et.on('test').take(1).forEach(e => {
  // This is fine.
  e.preventDefault();
  console.log('handled!');
})
.then(() => {
  console.log('forEach resolved as done');
});
  
console.log('before dispatch');
et.dispatchEvent(new Event('test'));
console.log('after dispatch');

// Logs
// "before dispatch"
// "handled!"
// "after dispatch"
// "forEach resolved as done"
```

domenic:
I strongly disagree with this direction, as previously discussed.

domfarolino:
If we want to really stick with the "Observable is a dual of Iterator" analogy, then I think we should keep the promise-returning methods as they are. `[1, 2, 3, 4].reduce((a, b) => { return a + b; })` doesn't return an iterator that, when exhausted, reduces to a final sum value. Instead, it eagerly "uses up" an iterator over the Iterable that it's called on, and immediately yields the final value. I think that'd be pretty clear with Observables too (hence the one-shot Promise return value).

As discussed a bit at TPAC, I don't think the concerns about the promise-returning methods interacting poorly with events is very serious. Consider `every()` and `some()` for example, which take a predicate callback whose return value feeds into an ultimate, single boolean.

 1. The point of the callback is to determine if an event passes some predicate test, not filter events or pass them along, so I don't think people will confuse these callbacks as the ideal opportunity to call `preventDefault()`. Although if they did, that's still fine since the callbacks run synchronously with respect to events that are emitted!
 2. The final value you get is a boolean, so you can't even expect to call `preventDefault()` on what the promise resolves to, so there's no footgun here.

There seems to be no reason to make those helpers return Observables, and making them return promises more-closely matches what Iterator helpers do.

[...]

PR #105, opened 2024-01-30
Spec the `forEach()` operator

domfarolino:
WPTs currently tracked in https://github.com/web-platform-tests/wpt/pull/42996/files#diff-b73e0c1acab4ab2faa26e0fd0c86eef3f96c535885d98e07b070b8b291e45de0, but will be merged into the Chromium prototype implementation CL, since I'll likely clean them up as I run the implementation against the tests.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -743,7 +743,45 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 <div algorithm>
   The <dfn for=Observable method><code>forEach(|callback|, |options|)</code></dfn> method steps are:
 
-    1. <span class=XXX>TODO: Spec this and use |callback| and |options|.</span>
+    1. Let |p| [=a new promise=].
+
+    1. If |options|'s {{SubscribeOptions/signal}} is not null:
+
+       1. If |options|'s {{SubscribeOptions/signal}} is [=AbortSignal/aborted=], then:
+
+          1. [=Reject=] |p| with |options|'s {{SubscribeOptions/signal}}'s [=AbortSignal/abort
+             reason=].
+
+          1. Return |p|.
+
+       1. [=AbortSignal/add|Add the following abort algorithm=] to |options|'s
+          {{SubscribeOptions/signal}}:
+
+          1. [=Reject=] |p| with |options|'s {{SubscribeOptions/signal}}'s [=AbortSignal/abort
+             reason=].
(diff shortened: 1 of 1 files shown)


domfarolino:
Feel free to take a look @keithamus, if you're interested. But no pressure.

spec.bs
```diff
@@ -743,7 +743,76 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 <div algorithm>
   The <dfn for=Observable method><code>forEach(|callback|, |options|)</code></dfn> method steps are:
 
-    1. <span class=XXX>TODO: Spec this and use |callback| and |options|.</span>
+    1. Let |p| [=a new promise=].
```
keithamus, line 746: This seems generally useful but I'm curious if there was any discussion weighing up why this should return a promise, vs undefined per `Array#forEach`, or if this should perhaps be renamed to differentiate itself from any symmetry of the array equivalent?

keithamus (approved):
This one LGTM 👍

spec.bs
domfarolino: I think the analogy is that this kind of since array iteration is synchronous and returns `undefined`, therefore Observable "iteration" (i.e., `forEach()`) would return a _`Promise<undefined>`_, since Observable "iteration" is asynchronous, and it still could be useful to know when said iteration/visiting is "done".

There is some discussion about this on https://github.com/WICG/observable/issues/69, but basically I'm not sure there is much utility in any operator returning undefined ever, vs _`Promise<undefined>`_ which carries strictly more information. But I'd be open to re-consideration if you want to strike up more conversation or file another issue.

keithamus: I don't want to rock the boat on what is likely well-thought out and a culmination of very long discussions, but having said that, part of me wonders if this is a little awkward and maybe it's just... not worth implementing `forEach` at all? I wonder how much value it adds. Presumably given how much of this is modeled on existing user-land libraries we can find some kind of usage figures to guide us on which API should be prioritized? Is `forEach` high in that list?

keithamus: I'm sure you have an answer for the above so I filed https://github.com/WICG/observable/issues/106 to hopefully document that!

domfarolino:
The Chromium implementation for this has landed with tests, so I'll merge this.

PR #110, opened 2024-02-08
Spec the `map()` operator

domfarolino:
The original version of this PR made it so if the `|mapper|` callback throws an error, it is "reported" to the global.

However, I've since fixed this to invoke the `error()` method on the relevant "Subscriber". This will either (a) [let the subscriber handle it][1], or (b) if there are no error steps, report the exception (due to the [default error steps][2]). That's the right behavior.

Additionally, in this error-case we *also* need to unsubscribe from the `|sourceObservable|`, which is accomplished by having the nexts steps abort an `AbortController` that `|signal|` is derived from, where `|signal|` is the `AbortSignal` passed into the outer Observable subscription.

Tests are in https://github.com/web-platform-tests/wpt/pull/42996/files#diff-8bcbcee0a9e290fe595f3a4d260648dc3900c67bab5925715011728b1df24d7f (written by @benlesh) but I will port them over to the Chromium CL, add more over there, and land them with the implementation. They'll get upstreamed to the external WPT repository with appropriate [Git commit attribution][3].

[1]: https://wicg.github.io/observable/#dom-subscriptionobserver-error
[2]: https://wicg.github.io/observable/#ref-for-default-error-algorithm
[3]: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/110.html" title="Last updated on Feb 8, 2024, 9:45 PM UTC (cb80f57)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/110/35b0376...cb80f57.html" title="Last updated on Feb 8, 2024, 9:45 PM UTC (cb80f57)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -664,6 +664,8 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
     1. Let |observable| be a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an
        algorithm that takes a {{Subscriber}} |subscriber| and does the following:
 
+       1. Let |controller| be a [=new=] {{AbortController}}.
+
        1. Let |sourceObserver| be a new [=internal observer=], initialized as follows:
 
           : [=internal observer/next steps=]
@@ -671,8 +673,8 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
                 |mappedValue| be the returned value.
 
                 If <a spec=webidl lt="an exception was thrown">an exception |E| was thrown</a>,
-                then run |subscriber|'s {{Subscriber/error()}} method, given |E|, and abort these
-                steps.
+                then run |subscriber|'s {{Subscriber/error()}} method, given |E|,
+                [=AbortController/signal abort=] |controller|, and abort these steps.
 
              1. Run |subscriber|'s {{Subscriber/next()}} method, given |mappedValue|.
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -659,7 +659,42 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
...
+
+    1. Let |observable| be a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an
+       algorithm that takes a {{Subscriber}} |subscriber| and does the following:
+
+       1. Let |controller| be a [=new=] {{AbortController}}.
+
+       1. Let |sourceObserver| be a new [=internal observer=], initialized as follows:
+
+          : [=internal observer/next steps=]
+          :: 1. [=Invoke=] |mapper| with the passed in <var ignore>value</var>, and let
```
benlesh, line 672: It should also supply an "index" (which is obviously a zero based counter). It doesn't need to supply a `source` argument like array does, because it really doesn't make sense for observable.

But `flatMap`, `map`, `filter`, `scan`, (or anything where there's a handler that takes a value, really) should probably be supplied an `index` that starts at 0 and increments for each value emitted.

benlesh (changes requested):

spec.bs
domfarolino: Done!

benlesh (approved):

Issue #114, opened 2024-02-10
Should add the checking subscriber's `active` state to the first step of `next(value)`/`error(error)`/`complete()` methods?


tetsuharuohzeki:
These supporting concepts does not check _subscriber_'s [_active_](https://wicg.github.io/observable/#subscriber-active) state.

- https://wicg.github.io/observable/#dom-subscriber-next
- https://wicg.github.io/observable/#dom-subscriber-error
- https://wicg.github.io/observable/#dom-subscriber-complete

And they use as null checking for each of _next/error/complete algorithm_ instead of checking _active_ concept, because of [close a subscription](https://wicg.github.io/observable/#close-a-subscription) concept set _active_ to `false` and set _next/error/complete algorithm_ to null.

However, I feel it's an implicit steps. I also think it's better to use _active_ concept and checking it at the first step of each of subscriber's _next(value)_ / _error(error)_ / _complete()_ concepts.

domfarolino:
Checking algorithm step null-ness vs checking `active` is not equivalent, unfortunately. You can have a `Subscriber` whose algorithms are null (for example, if nobody provided a [`complete()` handler](https://wicg.github.io/observable/#dom-subscriptionobserver-complete)) but is still active. So if `Subscriber#complete()` _only_ checked `active`, it will try and execute the complete steps, which may be null.

Does that make sense?

tetsuharuohzeki:
@domfarolino 

> So if Subscriber#complete() only checked active, it will try and execute the complete steps, which may be null.

By _[subscribe to an Observable](https://wicg.github.io/observable/#observable-subscribe-to-an-observable) concept_, the steps 2 lets a new _internal observer_ and it will be used to initialize _subscriber_ in the step 5.

> https://wicg.github.io/observable/#observable-subscribe-to-an-observable
> 
> 2. Let internal observer be a new [internal observer](https://wicg.github.io/observable/#internal-observer).
> ---
> 5. Let subscriber be a [new](https://webidl.spec.whatwg.org/#new) [Subscriber]> (https://wicg.github.io/observable/#subscriber), initialized as:
> 
>     - [next algorithm](https://wicg.github.io/observable/#subscriber-next-algorithm)
>         - internal observer’s [next steps](https://wicg.github.io/observable/#internal-observer-next-steps)
>      - [error algorithm](https://wicg.github.io/observable/#subscriber-error-algorithm)
[...]

PR #129, opened 2024-03-26
Add the step to check subscriber's active state to next(value)/error(error)/complete() as a shortcut

tetsuharuohzeki:
Fix https://github.com/WICG/observable/issues/114

***
<a href="https://pr-preview.s3.amazonaws.com/tetsuharuohzeki/observable/pull/129.html" title="Last updated on Jun 27, 2024, 8:21 PM UTC (9f8040d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:9f8040d.html" title="Last updated on Jun 27, 2024, 8:21 PM UTC (9f8040d)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -198,7 +198,7 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to 
 <div algorithm>
   The <dfn for=Subscriber method><code>next(|value|)</code></dfn> method steps are:
 
-    1. If [=this=]'s [=Subscriber/active=] is false, then return.
+    1. If [=this=]'s [=Subscriber/active=] is false, [=report the exception=] |error|, then return.
 
     1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
        Document=] is not [=Document/fully active=], then return.
--- a/spec.bs
+++ b/spec.bs
@@ -198,7 +198,7 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to 
 <div algorithm>
   The <dfn for=Subscriber method><code>next(|value|)</code></dfn> method steps are:
 
-    1. If [=this=]'s [=Subscriber/active=] is false, [=report the exception=] |error|, then return.
+    1. If [=this=]'s [=Subscriber/active=] is false, then return.
 
     1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
        Document=] is not [=Document/fully active=], then return.
--- a/spec.bs
+++ b/spec.bs
@@ -198,6 +198,8 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to 
 <div algorithm>
   The <dfn for=Subscriber method><code>next(|value|)</code></dfn> method steps are:
 
+    1. If [=this=]'s [=Subscriber/active=] is false, then return.
+
     1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
        Document=] is not [=Document/fully active=], then return.
 
@@ -220,6 +222,8 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to 
 <div algorithm>
   The <dfn for=Subscriber method><code>error(|error|)</code></dfn> method steps are:
 
+    1. If [=this=]'s [=Subscriber/active=] is false, then return.
+
     1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
        Document=] is not [=Document/fully active=], then return.
 
@@ -240,6 +244,8 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to 
 
 <div algorithm>
   The <dfn for=Subscriber method><code>complete()</code></dfn> method steps are:
+
+    1. If [=this=]'s [=Subscriber/active=] is false, then return.
 
     1. If [=this=]'s [=relevant global object=] is a {{Window}} object, and its [=associated
        Document=] is not [=Document/fully active=], then return.
--- a/spec.bs
+++ b/spec.bs
@@ -164,13 +164,13 @@ interface Subscriber {
 </xmp>
 
 Each {{Subscriber}} has a <dfn for=Subscriber>next algorithm</dfn>, which is a [=internal
-observer/next steps=]-or-null.
+observer/next steps=].
 
 Each {{Subscriber}} has a <dfn for=Subscriber>error algorithm</dfn>, which is an [=internal
-observer/error steps=]-or-null.
+observer/error steps=].
 
 Each {{Subscriber}} has a <dfn for=Subscriber>complete algorithm</dfn>, which is a [=internal
-observer/complete steps=]-or-null.
+observer/complete steps=].
 
 Each {{Subscriber}} has a <dfn for=Subscriber>teardown callbacks</dfn>, which is a [=list=] of
 {{VoidFunction}}s, initially empty.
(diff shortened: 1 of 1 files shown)


domfarolino:
Thanks! Do you think we can make this simplification go a bit further by making https://wicg.github.io/observable/#subscriber-next-algorithm, https://wicg.github.io/observable/#subscriber-error-algorithm, and https://wicg.github.io/observable/#subscriber-complete-algorithm all **non**-null, and then **not** clearing them in https://wicg.github.io/observable/#close-a-subscription?

spec.bs
```diff
@@ -210,6 +212,8 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to
 <div algorithm>
   The <dfn for=Subscriber method><code>error(|error|)</code></dfn> method steps are:
 
+    1. If [=this=]'s [=Subscriber/active=] is false, then return.
```
benlesh, line 215: This is correct, but there's an additional step. If the `Subscriber/active` is `false`, we'll want to report the `|error|` with `reportError`, because it can't be handled and we don't want it to be completely swallowed or ignored. cc @domfarolino

benlesh:
I'm not quite sure how the build is failing, maybe @domfarolino needs to look at it, or maybe it just needs rebased... however, it seems that @tetsuharuohzeki [you'll need to link your account to a W3C account](https://labs.w3.org/repo-manager/pr/id/WICG/observable/129) to pass on of the checks above.
image

domfarolino:
@benlesh I think your commit touched the wrong method `next()` instead of `error()`. Please revert/fix.

spec.bs
domfarolino: Right, that's what we do right now: https://wicg.github.io/observable/#dom-subscriber-error:~:text=Otherwise%20(i.e.%2C%20when%20error%20algorithm%20is%20null)%2C%20report%20the%20exception%20error

domfarolino: It's currently complicated in this PR because https://github.com/WICG/observable/pull/129/#issuecomment-2033200801 has not been addressed yet. If we address https://github.com/WICG/observable/pull/129/#issuecomment-2033200801, then we can use the `active` variable as the sole signal for whether we should proceed or not, instead of sometimes use `active` and sometime use "algorithm nullness".

benlesh:
@domfarolino oops, fixed.

tetsuharuohzeki:
> however, it seems that @tetsuharuohzeki [you'll need to link your account to a W3C account](https://labs.w3.org/repo-manager/pr/id/WICG/observable/129) to pass on of the checks above.
>
> https://github.com/WICG/observable/pull/129#issuecomment-2048472818

@benlesh 

I resolved that a moment ago.

----

> Do you think we can make this simplification go a bit further by making [wicg.github.io/observable/#subscriber-next-algorithm](https://wicg.github.io/observable/#subscriber-next-algorithm), [wicg.github.io/observable/#subscriber-error-algorithm](https://wicg.github.io/observable/#subscriber-error-algorithm), and [wicg.github.io/observable/#subscriber-complete-algorithm](https://wicg.github.io/observable/#subscriber-complete-algorithm) all non-null, and then not clearing them in [wicg.github.io/observable/#close-a-subscription](https://wicg.github.io/observable/#close-a-subscription)?
>
> https://github.com/WICG/observable/pull/129#issuecomment-2033200801

@domfarolino 

I would like to have a time to answer it to switch my brain to this spec :)
And I think it might be better to talk about it as the new issue.

domfarolino:
> And I think it might be better to talk about it as the new issue.

I think you are proposing to do this in a separate issue/follow-up, but I think this should be done in this PR. It is a simple change that is very inline with the changes this PR makes. I view the changes I've requested as a way to "finish up" the work started here, which would be great to do here!

benlesh:
The additions in this PR look correct to me.

It sounds like we're also discussing some related, but not exactly related, issue of converting a partial observer or next handler to a Subscriber, in which case if no error handler is provided, a `reportError` is put in its place. I thought we were already doing that?

In any case, IMO, this PR is good to stand on its own. What do you think, @domfarolino ?

domfarolino:
> It sounds like we're also discussing some related, but not exactly related, issue of converting a partial observer or next handler to a Subscriber, in which case if no error handler is provided, a `reportError` is put in its place. I thought we were already doing that?

I don't think so? I'm not sure what a partial observer is, or what you mean about the error handling stuff. The only thing I'm discussing / requesting in https://github.com/WICG/observable/pull/129#issuecomment-2033200801 and https://github.com/WICG/observable/pull/129#discussion_r1560113116 is that we fully migrate away from the old world of making ["next algorithm"](https://wicg.github.io/observable/#subscriber-next-algorithm), "error algorithm", and "complete algorithm" signals for inactivity, by no longer making them nullable, and no longer nulling them out in the "Close" algorithm. I think that would just make it more likely that we would remember to not use those as signals for inactivity in future PRs, and instead solely use the "active" boolean as this PR starts to do.

tetsuharuohzeki:
I'm sorry about no response for this issue.

I'm working today to make _Subscriber_ concept's next/error/complete algorithm non-null as @domfarolino's suggestion. I seem it's possible but I also checking it's valid again.

domfarolino (commented):
Note that in https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:f80ab90.html#dom-subscriber-error and https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:f80ab90.html#dom-subscriber-complete, we no longer have to pull out the error steps or complete steps into a separate variable before calling "Close", because the close algorithm no longer sets those steps to null. So now we can just call "close" and then use `this's error steps` (same for complete) directly. Let me know if that doesn't make sense.

tetsuharuohzeki:
@domfarolino 

> Let me know if that doesn't make sense.

I concern that your suggestion makes to allow infinite recursion accidentally.

The current spec calls error steps or complete steps **after** "close subscription" (set active=`false`) and run signal abort to stop a subscription chain completely.

But your suggestion is that calling error steps or complete steps **before** "close subscription" and "run signal abort". That does not wait to stop a subscription chain.

I fear your suggestion that may introduce a chance probably for accidental infinite recursion. It should be discussed in a new topic.

domfarolino:
> But your suggestion is that calling error steps or complete steps before "close subscription" and "run signal abort".

No my suggestion is still calling the error steps or complete steps after Close. I just think we can get rid of these lines:

1. https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:f80ab90.html#dom-subscriber-next:~:text=Let%20error%20algorithm%20be%20this%20%27s%20error%20algorithm
2. https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:f80ab90.html#dom-subscriber-next:~:text=Let%20complete%20algorithm%20be%20this%20%27s%20complete%20algorithm

tetsuharuohzeki:
> > But your suggestion is that calling error steps or complete steps before "close subscription" and "run signal abort".
> 
> No my suggestion is still calling the error steps or complete steps after Close. I just think we can get rid of these lines:
> 
>     1. https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:f80ab90.html#dom-subscriber-next:~:text=Let%20error%20algorithm%20be%20this%20%27s%20error%20algorithm
> 
>     2. https://pr-preview.s3.amazonaws.com/WICG/observable/129/98a4f07...tetsuharuohzeki:f80ab90.html#dom-subscriber-next:~:text=Let%20complete%20algorithm%20be%20this%20%27s%20complete%20algorithm

Ah, that's make sense. Okay. I'll address them.

tetsuharuohzeki:
I pushed the addressed commit.

domfarolino (approved):

Issue #52, opened 2023-08-30
Include `switchMap` operator?


Yogu:
The explainer includes `flatMap`, but not `switchMap`. I think `switchMap` should be included because it is essential in a common pattern:

- given a source (e.g. the value of an input)
- execute an asynchronous action when the source emits
- and if the previous execution has not completed yet, cancel it.

A simple example is a search bar that triggers a network request, but should cancel the previous one when the user types too quickly.

I don't think I have ever used the rxjs `flatMap` operator, but I'm using `switchMap` for the pattern describe above all the time.

---

From @benlesh: I created a duplicate issue (#90) for this accidentally, so I'll include the use cases and text I posted there here:

`switchMap` was mentioned at TPAC as being an API of interest. It's a method that is especially useful for composing events in the browser because of its inherent cancellation. If API bloat is a concern, I would gladly remove `toArray()`, `some()`, `find()` et al, as those are minimally useful with observables.

Use cases are things like:

**Use Case: Lookahead Search**

In the below scenario, the result of the `fetch` would simply be ignored.

It's plausible that we could provide a signal along with the `switchMap` 
callback, as well. In fact, I'd almost encourage that, as it might make fetch
more useful. But I think we'd want to discuss the ergonomics there, and that could
be debated and added later. (For example, we might want to have it automatically
swallow DOMExceptions named `AbortError`?)

```ts
const input = document.getElementById('search');
const resultList = document.getElementById('result-list');

input.on('input')
  .switchMap(async (e) => {
    const response = await fetch(`/search?q={e.target.value}`);

    if (!response.ok) {
      console.warn(`Search response error: ${response.status} ${response.statusText}`
      return;
    }
  
    return response.json();
  })
  .subscribe((results) => {
    resultList.innerHTML = results
      .map(result => `<li><a href="${result.href}">${result.text}</a></li>`)
      .join('')
  })
```

**Use Case: Changing Connections**

Below, every time a URL is changed in an URL input, if it's a valid URL,
it will disconnect from the previous web socket and connect to a new one.

```ts
function streamData(url) {
  return new Observable(subscriber => {
    const ws = new WebSocket(url);
    ws.onmessage = e => subscriber.next(e);
    subscriber.addTeardown(() => ws.close());
  })
}

const urlInput = document.getElementById('url-input');

urlInput.on('input')
  .filter(e => e.target.checkValidity())
  .switchMap(e => streamData(e.target.value))
  .subscribe(console.log);
```

[...]

Issue #126, opened 2024-03-05
Initial methods to add, and a few to remove


benlesh:
### Summary

**Adding**
1. `Observable switchMap(mapper)`
2. `Observable catch(mapper)`
3. `Observable finally(callback)`
4. `Promise first()`
5. `Promise last()`
6. `Observable scan(reducer, initialValue?)`

**Removing**
1. `Promise some(predicate)`
2. `Promise every(predicate)`
3. `Promise find(predicate)`
4. `Promise toArray()`
5. `Promise reduce(reducer, initialValue?)`

**Desire to followup later**
1. `Observable tap(callback/Observer)`

#### `reduce` and `toArray()` don't make sense for events, `scan` does

While this does deviate from the async iterator helpers proposal, observables target a different set of use cases. `reduce()` and `toArray()` in particular make much more sense in an async iterable, where you might be iterating through an array of items, then performing some async action to get a new array of items; Where it doesn't make much sense to reduce mouse clicks or make an array of move events.

To that end, `scan` is a much more common task, where one might want to continuously emit reduced state (ala Redux or React's `useReducer`) as it's changed by events, and `scan().last()` is the same as `reduce()`. Similarly, in my many years working with observables via RxJS, I've found that `toArray()` is usually adjacent to poorly written code, that does something like take an observable of arrays, flatten it to single values, do something with them, then re-accumulate them as an array. An iterable or simple array methods would likely be a better choice for such a task.

#### `catch` and `finally`

It may be obvious, but iterators and async iterators simply have a different mechanism for communicating errors or finalization and handling them. Observable needs `catch` and `finally` methods that behave very similarly to the ones that promise provides.

`catch` has a common use case with `flatMap` for keeping source streams alive when child streams error:

```ts
// Without the catch statement below,
// ALL observation would stop if the observable
// returned by `startDataStream()` errored.

element.on('click').flatMap(() => {
  return startDataStream()
    .catch(err => {
      console.error(err);
      return []; // empty
    })
})
.subscribe(console.log);
```

`finally` has use cases around clean up that you want to occur on unsubscription, completion, or error, and you reproduce between subscriptions to the same observable.

```ts
// Without the `finally` method use below,
// the `socket` will not be cleaned up appropriately

const controller = new AbortController();
const { signal } = controller;

takeThreeButton.on('click')
  .flatMap(() => {
    const socket = new WebSocket(url);
    return socket.on('message')
      .take(3)
      .finally(() => socket.close())
  })
[...]

PR #130, opened 2024-03-29
Spec the `switchMap` operator

domfarolino:
This PR spec's the `switchMap()` operator.

This PR closes https://github.com/WICG/observable/issues/52 and makes progress on https://github.com/WICG/observable/issues/126.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/130.html" title="Last updated on Apr 23, 2024, 3:40 PM UTC (c4db0e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/130/e3e8f94...c4db0e1.html" title="Last updated on Apr 23, 2024, 3:40 PM UTC (c4db0e1)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1014,7 +1014,7 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
                 Note: If |activeInnerAbortController| is not null, then we don't immediately
                 complete |subscriber|. Instead, the [=switchmap process next value steps=] will
                 complete |subscriber| when the inner subscription finally becomes inactive on its
-                own.
+                own volition.
 
              1. If |activeInnerAbortController| is null, run |subscriber|'s
                 {{Subscriber/complete()}} method.
@@ -1031,7 +1031,8 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 <div algorithm>
   The <dfn>switchmap process next value steps</dfn>, given an {{any}} |value|, a {{Subscriber}}
   |subscriber|, a {{Mapper}} |mapper|, and <b>references</b> to all of the following: an
-  {{AbortController}} |activeInnerAbortController|, and a [=boolean=] |outerSubscriptionHasCompleted|:
+  {{AbortController}} |activeInnerAbortController|, and a [=boolean=]
+  |outerSubscriptionHasCompleted|, are to run these steps:
 
     1. Let |mappedResult| be the result of [=invoking=] |mapper| with |value|.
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -958,6 +959,125 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
...
+    1. Return |observable|.
+</div>
+
+<div algorithm>
+  The <dfn>switchmap process next value steps</dfn>, given an {{any}} |value|, a {{Subscriber}}
+  |subscriber|, a {{Mapper}} |mapper|, and <b>references</b> to all of the following: an
+  {{AbortController}} |activeInnerAbortController|, and a [=boolean=]
+  |outerSubscriptionHasCompleted|, are to run these steps:
+
+    1. Let |mappedResult| be the result of [=invoking=] |mapper| with |value|.
```
benlesh, line 1037: With a value and index. All of the mapping functions should take an index as a second argument. Just like `array.map((value, index) => 'result')`.

spec.bs
```diff
@@ -958,6 +959,125 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
...
+          :: Run |subscriber|'s {{Subscriber/error()}} method, given the passed in <var
+             ignore>error</var>.
+
+          : <dt id=switchmap-complete-steps>[=internal observer/complete steps=]</dt>
+          :: 1. Set |outerSubscriptionHasCompleted| to true.
+
+                Note: If |activeInnerAbortController| is not null, then we don't immediately
+                complete |subscriber|. Instead, the [=switchmap process next value steps=] will
+                complete |subscriber| when the inner subscription finally becomes inactive on its
+                own volition.
```
benlesh, line 1017: It will only complete if the inner subscriber completes. "inactive" almost implies that if it waits a while or maybe errors, it will still complete. Both the outer source and the inner source, if there is one, have to complete in order for the subscriber to complete.

benlesh (changes requested):
Mostly okay, just a couple of things that I caught.

spec.bs
domfarolino: Looks like `flatMap()` is also missing this (although the implementation works). Filed https://github.com/WICG/observable/issues/139 for that. I'll fix `switchMap()` here.

domfarolino: Done.

spec.bs
domfarolino: Improved this a bit. Done.

domfarolino:
I think everything is addressed here so I am going to land this.

Issue #132, opened 2024-04-02
Should `first()` and `last()` support a default value?


domfarolino:
Just as an initial note to this issue: resolution of this issue is not required for the shipping of either `first()` or `last()` APIs.

It came to my attention through looking at the [original WPTs for these APIs](https://github.com/web-platform-tests/wpt/pull/44486) that there was an expectation of a "default" value for these APIs, in case they complete without emitting any values. RxJS appears to support this too (see https://rxjs.dev/api/operators/first & https://rxjs.dev/api/operators/last).

Personally I don't see this as necessary, since you could easily catch the case where completion happens without any values being emitted, and handle it yourself. But maybe the ergonomics is useful enough to include this? Either way, support for a default value could come at any point in the future too I think.

benlesh:
The analog in RxJS (because promises) is actually `firstValueFrom` and `lastValueFrom`, which do support default values. As if they're empty they should reject, and that's not always desirable.

IMO, I think it's a strong add, and anecdotally, I've seen it used frequently

domenic:
Back when I did .NET, their enumerables had LastOrDefault and FirstOrDefault, which I thought were clear names and I remember using them often.

I do feel wary about getting ahead of iterator helpers for things like this, though.

bakkot:
Iterator helpers are unlikely to get `first` or `last` methods at all, IMO. They don't seem clearly motivated to me for iterators and I don't think anyone's expressed interest in pushing for them.

domenic:
> Do you mean "getting ahead" by including these methods at all, in the first place?

Yes. Although if @bakkot thinks we're just not going to have this family for iterators, then maybe it's fine. I think in some ecosystems we'd push for consistency by including even useless methods, as long as they make sense. But oh well.

> so changing the name to `firstValueFrom` or `firstOrDefault` feels kinda clunky for most really simple, cute use cases

Sorry, my suggestion was four methods: `first()`, `firstOrDefault(defaultValue)`, `last()`, `lastOrDefault(defaultValue)`. In JS you actually get a nice bonus where `firstOrDefault()` with no argument is equivalent to `firstOrDefault(undefined)`, which seems helpful.

domenic:
I've thought about this a slight bit more and wanted to capture my thoughts.

Potential functionality you might want:

- first/last/find, or `undefined` if empty/not found
- first/last/find, or user-supplied default if empty/not found
- first/last/find, or throw if not found

[...]

PR #131, opened 2024-04-02
Spec the `first()` operator

domfarolino:
Closes https://github.com/WICG/observable/issues/65 and makes progress on https://github.com/WICG/observable/issues/126.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/131.html" title="Last updated on May 5, 2024, 2:42 PM UTC (501b4f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/131/e3e8f94...501b4f7.html" title="Last updated on May 5, 2024, 2:42 PM UTC (501b4f7)">Diff</a>

PR #137, opened 2024-04-08
Spec `some()`, `every()`, and `find()` operators

domfarolino:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/137.html" title="Last updated on Apr 23, 2024, 7:23 PM UTC (d609f27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/137/b4b9d43...d609f27.html" title="Last updated on Apr 23, 2024, 7:23 PM UTC (d609f27)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1116,19 +1116,148 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 <div algorithm>
   The <dfn for=Observable method><code>every(|predicate|, |options|)</code></dfn> method steps are:
 
-    1. <span class=XXX>TODO: Spec this and use |predicate| and |options|.</span>
+    1. Let |p| [=a new promise=].
+
+    1. Let |controller| be a [=new=] {{AbortController}}.
+
+    1. Let |internal options| be a new {{SubscribeOptions}} whose {{SubscribeOptions/signal}} is the
+       result of [=creating a dependent abort signal=] from the list
+       «|controller|'s [=AbortController/signal=], |options|'s
+       {{SubscribeOptions/signal}} if non-null», using {{AbortSignal}}, and the [=current realm=].
+
+    1. If |internal options|'s {{SubscribeOptions/signal}} is [=AbortSignal/aborted=], then:
+
+       1. [=Reject=] |p| with |internal options|'s {{SubscribeOptions/signal}}'s
+          [=AbortSignal/abort reason=].
+
+       1. Return |p|.
+
+    1. [=AbortSignal/add|Add the following abort algorithm=] to |internal options|'s
+       {{SubscribeOptions/signal}}:
+
+       1. [=Reject=] |p| with |internal options|'s {{SubscribeOptions/signal}}'s [=AbortSignal/abort
+          reason=].
+
+    1. Let |observer| be a new [=internal observer=], initialized as follows:
+
+       : [=internal observer/next steps=]
+       :: 1. [=Invoke=] |predicate| with the passed in <var ignore>value</var>, and let |passed| be
+             the returned value.
+
+             If <a spec=webidl lt="an exception was thrown">an exception |E| was thrown</a>, then
+             [=reject=] |p| with |E|, and [=AbortController/signal abort=] |controller| with |E|.
+
+         1. If |passed| is false, then [=resolve=] |p| with false, and [=AbortController/signal
+            abort=] |controller|.
+
+       : [=internal observer/error steps=]
+       :: [=Reject=] |p| with the passed in <var ignore>error</var>.
+
+       : [=internal observer/complete steps=]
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1112,19 +1112,148 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
...
+             [=reject=] |p| with |E|, and [=AbortController/signal abort=] |controller| with |E|.
+
+         1. If |passed| is true, then [=resolve=] |p| with |value|, and [=AbortController/signal
+            abort=] |controller|.
+
+       : [=internal observer/error steps=]
+       :: [=Reject=] |p| with the passed in <var ignore>error</var>.
+
+       : [=internal observer/complete steps=]
+       :: [=Resolve=] |p| with false.
```
benlesh, line 1202: This should reject (With a `RangeError` maybe, to follow suit with `first` and `last`) because the value was not found. What if the consumer is trying to find `false`? `observable.find(v => v === false)`? 

In fact, there should probably be a default option to override that (as there probably should be for `first` and `last`): `observable.find(predicate, { defaultValue: 'Did not find the value' })`.

benlesh (changes requested):

spec.bs
bakkot: The default behavior should really be to match arrays / iterators, which return `undefined` in the case that a match was not found.

(Which, yes, makes it hard to search for `undefined`. But I don't think this should break with precedent)

domfarolino: Done! Hmm, in that case I'm thinking `first()` and `last()` should probably do the same then, instead of using `RangeError`s, even if [Iterator helpers is unlikely to get those operators](https://github.com/WICG/observable/issues/132#issuecomment-2038410116).

domfarolino: (I've left a comment about this on https://github.com/WICG/observable/pull/131#issuecomment-2083248807 for `first()` and `last()`.)

domfarolino:
Everything should be addressed here. I think I'll land this at the end of today if you want to take another look @benlesh.

spec.bs
benlesh: **TL;DR: I can concede this if we think it makes everything more palatable for the committee. I do strongly advise we allow a `defaultValue` configuration though, to cover the corner case**
 
@domfarolino @bakkot FWIW... the behavior of RxJS 1 - 6's `Observable.prototype.toPromise()` for about 10 years was to emit `undefined` in cases where an observable completed without emitting a value.

Over the years, this caused enough issues and small bugs that it was agreed that `firstValueFrom(observable)` and `lastValueFrom(observable)` was to throw an `EmptyError` (specific to RxJS), unless a `defaultValue` was specified.

`toPromise()`, `firstValueFrom`, and `lastValueFrom` are _heavily used_ as a means of interop with promise-based APIs and async/await.

On the other hand, RxJS's `find` operator, does indeed emit `undefined` if it can't find what you're looking for. This has largely been okay and not caused issues, however, `find` is _rarely_ used from what I've seen over the years. We do have a `first(predicate, defaultValue?)` operator that will throw an `EmptyError` if it doesn't find the value, and no default was provided.

Now... all of this said, the major issue with RxJS's `toPromise()` from what I'm seeing through all filed issues, was the fact that (since RxJS 5 was originally written against TypeScript 1.8) the return type of `toPromise()` was wrong, being `Promise<T>` instead of `Promise<T | undefined>`, and the name was too general, as it always gave you the last value of the observable. And given that people largely wanted `Promise<T>`, and we also wanted to remove the method from `Observable`, we changed to have it reject with an `EmptyError` and named it more appropriately, as `lastValueFrom()`. Then we added `firstValueFrom()` because it was common pattern for people to do `observable.take(1).toPromise()`. 

I'm totally okay with moving back to having it return `undefined` if the item isn't found, or nothing is emitted, if we think that helps move things along. But I still recommend a configurable default value. I'm not sure if there's prior art for that though.

benlesh: Also: Apologies, I merged #133 before I saw this thread.

bakkot: I'm not as attached to returning `undefined` for `first` and `last`, though it does seem more consistent/usable than an error would be. Mostly I want `find` specifically to match arrays/iterators, since it exists under that name already, whereas `first` and `last` don't.

I agree that in theory a configurable fallback value for `find` (and the other two) would be nice, though in practice I can't think of a time that's ever actually come up for me.

domfarolino: Yeah +1 to pretty much all of that. I'll work on making #131 #144 not use `RangeError`, instead yielding `undefined` in the vacuous case, and then we have https://github.com/WICG/observable/issues/132 still open which I think is important but not a hard blocker for initially launching.

PR #141, opened 2024-04-23
Fix predicate. Fix #140.

### LINKED ISSUE #140: Predicate function should include an index parameter
In JavaScript land, [`filter()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/filter#parameters), [`every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/every#parameters), [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/some#parameters), and [`find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/find#parameters) all have an index-consuming Predicate function.

This is in contrast with the Observable API's [Predicate functions](https://wicg.github.io/observable/#callbackdef-predicate) which does not currently consume an index parameter. I should fix this and update the only spec'd method that uses `Predicate` right now, which is `filter()`. This ideally should be done before https://github.com/WICG/observable/pull/137 lands.

domfarolino:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/141.html" title="Last updated on Apr 23, 2024, 7:14 PM UTC (b218cd7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/141/5fac9fd...b218cd7.html" title="Last updated on Apr 23, 2024, 7:14 PM UTC (b218cd7)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -337,7 +337,7 @@ dictionary SubscribeOptions {
   AbortSignal signal;
 };
 
-callback Predicate = boolean (any value);
+callback Predicate = boolean (any value, unsigned long long index);
 callback Reducer = any (any accumulator, any currentValue);
 callback Mapper = any (any element, unsigned long long index);
 // Differs from Mapper only in return type, since this callback is exclusively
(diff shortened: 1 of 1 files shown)


domfarolino (commented):
This is a straightforward fix and the implementation agrees, so I will land this.

Issue #151, opened 2024-06-26
`finally()` operator timing semantics


domfarolino:
TL;DR: On unsubscription (i.e., `abort()`), what should the timing of the `finally()` operator's callback be with respect to the source Observable's teardowns?

When writing the spec for, and implementing, the currently-unspecified [`finally()` operator](https://wicg.github.io/observable/#dom-observable-finally), I was reviewing the proposed web platform tests and had a comment about this bit: https://github.com/web-platform-tests/wpt/pull/44482/files#diff-b1a6cf94fa2c551227215a7556fbc3114006fca82549ba48a27f3dafe2564f01R186. This made me think more about the timing of the `finally()` operator's callback with respect to the source Observable's teardowns.

## Background: unsubscription teardown semantics

When you subscribe (with an `AbortSignal`) to an Observable returned by an operator, two things happen:

1. A dependent signal [is created](https://wicg.github.io/observable/#ref-for-subscriber-signal%E2%91%A0) from the signal that was passed into `subscribe()`.
2. The intermediate Observable is subscribed to with that dependent signal ([example](https://wicg.github.io/observable/#ref-for-observable-subscribe-callback%E2%91%A5)), and its native subscribe callback runs.
3. The operator subscribes to _its_ upstream Observable with the dependent signal ([example](https://wicg.github.io/observable/#ref-for-observable-subscribe-to-an-observable%E2%91%A6)). This dependent signal goes through (1) above and gets turned into a new dependent signal before subscription, of course.

So in the following code:

```js
const source = new Observable(subscriber => ...);

const ac = new AbortController();
source.take(1).subscribe({}, {signal: ac.signal});
```

There are three `AbortSignals` involved:

1. The parent one: `ac.signal`.
2. A dependent of (1), for `take()`s Subscriber's `Subscriber#signal`. (It is purely internal and not web accessible).
3. A dependent of (2), exposed in `subscriber` above.

If you call `ac.abort()`, then the abort signals are aborted in the order they're listed above. That's because DOM defines that a parent signal gets fully aborted before any of its dependents: https://dom.spec.whatwg.org/#abortsignal-signal-abort. Practically speaking, this means if you added an `abort` event listener to `ac.signal` and `subscriber.signal`, the `ac.signal` one would fire first, and the `subscriber.signal` one would fire last.

## Integrating `finally()`

Now consider the requirement that the `finally()` callback needs to be called when a consumer unsubscribes from an Observable:

```js
const ac = new AbortController();

[...]

PR #153, opened 2024-07-01
Spec the `finally()` operator

domfarolino:
This PR specs the `finally()` operator, registering the callback to its internal Subscriber's teardown list.

~Mirroring the `finally` keyword, its semantics are such that on both producer-initiated and consumer-initiated unsubscription, the finally operator's callback runs. The only remaining open question about this operator has to do with subtle timing between the finally() operator and source Observable teardown execution. See https://github.com/WICG/observable/issues/151 for discussion there. I don't think that needs to block this PR, but it does need to resolve one way or another.~

The above issues are struckthrough and obsolete; see https://github.com/WICG/observable/issues/151#issuecomment-2675072493.

Tests are being landed in https://chromium-review.googlesource.com/c/chromium/src/+/5654720.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/153.html" title="Last updated on Feb 21, 2025, 6:11 PM UTC (da2f460)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/153/a091c5a...da2f460.html" title="Last updated on Feb 21, 2025, 6:11 PM UTC (da2f460)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1545,7 +1545,87 @@ For now, see
 <div algorithm>
   The <dfn for=Observable method><code>finally(|callback|)</code></dfn> method steps are:
 
-    1. <span class=XXX>TODO: Spec this and use |callback|.</span>
+    1. Let |sourceObservable| be [=this=].
+
+    1. Let |observable| be a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an
+       algorithm that takes a {{Subscriber}} |subscriber| and does the following:
+
+       1. Let |finally callback steps| be the following steps:
+
+         1. [=Invoke=] |callback|.
+
+            If <a spec=webidl lt="an exception was thrown">an exception |E| was thrown</a>, then run
+            |subscriber|'s {{Subscriber/error()}} method, given |E|, and abort these steps.
+
+       1. [=AbortSignal/add|Add the algorithm=] |finally callback steps| to |subscriber|'s
+          [=Subscriber/signal=].
+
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1545,7 +1545,33 @@ For now, see
...
 
-    1. <span class=XXX>TODO: Spec this and use |callback|.</span>
+    1. Let |sourceObservable| be [=this=].
+
+    1. Let |observable| be a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an
+       algorithm that takes a {{Subscriber}} |subscriber| and does the following:
+
+       1. Run |subscriber|'s {{Subscriber/addTeardown()}} method with |callback|.
+
+       1. Let |sourceObserver| be a new [=internal observer=], initialized as follows:
```
benlesh, line 1555: You can just pass the `subscriber` directly to the `subscribe`. There's no need to create another reference in this one-off case.

```ts
Observable.prototype.finally = (callback) => {
  return new Observable(subscriber => {
    subscriber.addTeardown(callback)
    this.subscribe(subscriber, { signal: subscriber.signal })
  })
}
```

benlesh (changes requested):

spec.bs
benlesh: Oops, I forgot you're in a differently typed environment. I retract my comment. It's fine.

benlesh (approved):

PR #154, opened 2024-07-09
Don't use dependent abort signals inside `Subscriber`

domfarolino:
Before this pull request, a `Subscriber`'s internal signal — the one exposed by [`Subscriber#signal()`][1] — was a [dependent signal][3] based on two input signals:

1. The [SubscribeOptions#signal][2], if non-null; and
2. The Subscriber's internal AbortController's (named `complete or error controller` before this CL) AbortSignal.

Then, whenever the internal _dependent_ signal is aborted, we run the **abort algorithm** which (**a**) "closes" the subscription (sets `subscriber.active` to false), and (**b**) executes all teardowns.

But (**1**) the use of _dependent_ AbortSignals (i.e., making each Subscriber's signal dependent on its downstream Subscriber's signal), and (**2**) tying a Subscriber's teardown execution to its internal signal's abort algorithms, introduced inconsistent teardown ordering for Observables chained together. Specifically, the order of teardown execution _across_ Observables in a chain depended on whether the Observable chain experienced producer-initiated vs. consumer-initiated
unsubscription. Example:

```js
const a = new Observable(subscriber => {
  window.sourceSubscriber = subscriber;
  subscriber.addTeardown(() => console.log('A teardown'));
});

const b = new Observable(subscriber => {
  subscriber.addTeardown(() => console.log('B teardown'));
  a.subscribe({complete: () => subscriber.complete()}, {signal: subscriber.signal});
})

const signal = ac.signal;
b.subscribe({}, {signal});

// Prints `B teardown`, then `A teardown`
// ac.abort();

// Prints `A teardown`, then `B teardown`
// window.sourceSubscriber.complete();
```

Upstream teardowns (`a`) should run before downstream teardowns (`b`). The reason the consumer-initiated case is backwards is due to the ordering or abortion among dependent AbortSignals. Specifically, once we abort the developer passed-in AbortSignal, the first thing that runs are the **abort algorithms**, which [runs local teardowns first][4]. Then, DOM [proceeds abort any dependent (aka upstream) signals][5], which _then_ runs those upstream teardowns, before aborting even more upstream signals, and so on.

----

To fix this, this PR introduces two small but significant changes to the underlying AbortSignal infrastructure in Observables.

The first thing this PR does is stop making a Subscriber's internal signal a _dependent signal_ based on the developer-supplied one (if one exists). The abort timing of dependent signals led to downstream Subscribers being aborted (teardowns and all) before upstream Subscribers (in the consumer-initiated unsubscription case, that is).

Instead, now the Subscriber maintains an internal AbortController (similar to today's "complete or error controller"), and its internal signal is that controller's signal. Then, like before, if a [signal is provided in SubscribeOptions][6], we add an abort algorithm to it to [**close the subscription**][7].

The second change this PR introduces is a change to the "close" algorithm itself. In addition to just setting `active` to false, it now aborts the Subscriber's internal controller. This kicks off the aborting of any upstream Subscribers right away, since those upstream "close" algorithms run (as an abort algorithm) in response to the downstream Subscriber's signal abort. This means "close" climbs all the way to the top-most upstream Subscriber/AbortSignal right away, before running any teardowns. Only after a subscription has been "closed", and after all upstream signals have been aborted, does the current Subscriber run its teardowns.

This changes the timing of things. In the consumer-initiated unsubscription case, instead of downstream signals running their teardowns and then "closing" their dependent (upstream) signals after, the first thing a downstream signal does in response to abort is "close" any upstream subscribers first.

Now imagine you have the following Observable chain:

```
a
  .operator() <-- (b) Observable
  .operator() <-- (c) Observable
  .subscribe({}, {signal});
```

Upon aborting `signal` you'd get the following flow:

```
1. Downstream signal aborted
  1. (C) "Close" algorithm runs
    1. (C) Aborts its internal controller
      1. (B) "Close" algorithm runs
        1. (B) Aborts its internal controller
          1. (A) "Close" algorithm runs
            1. (A) Aborts its internal controller
            2. (A) Teardowns run
        2. (B) Teardowns run
    2. (C) Teardowns run
```

**Before** this PR, the timing flow was like this:

```
1. Downstream signal aborted
  1. (C) Signal aborted
    1. (C) Close algorithm
      1. (C) Teardowns run
    2. (B) Signal aborted
      1. (B) Close algorithm
        1. (B) Teardowns run
      2. (A) Signal aborted
        1. (A) Close algorithm
          2. (A) Teardowns run
```

[1]: https://wicg.github.io/observable/#dom-subscriber-signal
[2]: https://wicg.github.io/observable/#dom-subscribeoptions-signal
[3]: https://dom.spec.whatwg.org/#create-a-dependent-abort-signal
[4]: https://dom.spec.whatwg.org/#abortsignal-signal-abort:~:text=For%20each%20algorithm%20of%20signal%E2%80%99s%20abort%20algorithms%3A%20run%20algorithm
[5]: https://dom.spec.whatwg.org/#abortsignal-signal-abort:~:text=For%20each%20dependentSignal%20of%20signal%E2%80%99s%20dependent%20signals%2C%20signal%20abort%20on%20dependentSignal%20with%20signal%E2%80%99s%20abort%20reason
[6]: https://wicg.github.io/observable/#dom-subscribeoptions-signal
[7]: https://wicg.github.io/observable/#close-a-subscription

Tests are being written and landed in https://chromium-review.googlesource.com/c/chromium/src/+/5676226.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/154.html" title="Last updated on Jul 9, 2024, 7:53 PM UTC (f1c5e86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/154/7b4cb07...f1c5e86.html" title="Last updated on Jul 9, 2024, 7:53 PM UTC (f1c5e86)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -36,7 +36,7 @@ urlPrefix: https://dom.spec.whatwg.org; spec: DOM
       text: signal; url: event-listener-signal
     for: AbortSignal
       text: dependent signals; url: abortsignal-dependent-signals
-      text: signal abort
+      text: signal abort; url:abortsignal-signal-abort
 </pre>
 
 <style>
@@ -175,14 +175,8 @@ observer/complete steps=].
 Each {{Subscriber}} has a <dfn for=Subscriber>teardown callbacks</dfn>, which is a [=list=] of
 {{VoidFunction}}s, initially empty.
 
-Each {{Subscriber}} has a <dfn for=Subscriber>complete or error controller</dfn>, which is an
+Each {{Subscriber}} has a <dfn for=Subscriber>subscription controller</dfn>, which is an
 {{AbortController}}.
-
-Each {{Subscriber}} has a <dfn for=Subscriber>signal</dfn>, which is an {{AbortSignal}}.
-
-Note: This is a [=create a dependent abort signal|dependent signal=], derived from both
-[=Subscriber/complete or error controller=]'s [=AbortController/signal=], and
-{{SubscribeOptions}}'s {{SubscribeOptions/signal}} (if non-null).
 
 Each {{Subscriber}} has a <dfn for=Subscriber>active</dfn> boolean, initially true.
 
@@ -193,7 +187,7 @@ The <dfn attribute for=Subscriber><code>active</code></dfn> getter steps are to 
 [=Subscriber/active=] boolean.
 
 The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to return [=this=]'s
-[=Subscriber/signal=].
+[=Subscriber/subscription controller=]'s [=AbortController/signal=].
 
 <div algorithm>
   The <dfn for=Subscriber method><code>next(|value|)</code></dfn> method steps are:
@@ -228,8 +222,6 @@ The <dfn attribute for=Subscriber><code>signal</code></dfn> getter steps are to 
 
     1. [=close a subscription|Close=] [=this=].
 
-    1. [=AbortController/Signal abort=] [=this=]'s [=Subscriber/complete or error controller=].
-
     1. Run [=this=]'s [=Subscriber/error algorithm=] given |error|.
 
        [=Assert=]: No <a spec=webidl lt="an exception was thrown">exception was thrown</a>.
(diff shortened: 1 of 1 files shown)


domfarolino:
Ben has approved the tests, and the Chromium implementation updates for this have landed, so I'll merge the spec now.

PR #155, opened 2024-07-11
Make take/drop not share state across subscriptions

Jamesernator:
As spec-ed the behaviour of take/drop means this prints the values only once:

```js
const obs = Observable.from([1,2,3,4,5,6]);
const firstThree = obs.take(3);

firstThree.forEach((v) => {
    console.log(v);
});

// This doesn't print anything
firstThree.forEach((v) => {
    console.log(v);
});
```

The intended behaviour I would expect is that each subscription has it's own counter.

The fix provided here just makes a new variable local to each subscription rather than re-using the argument to take/drop.

***
<a href="https://pr-preview.s3.amazonaws.com/Jamesernator/observable/pull/155.html" title="Last updated on Jul 11, 2024, 4:53 AM UTC (0316480)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/155/7b4cb07...Jamesernator:0316480.html" title="Last updated on Jul 11, 2024, 4:53 AM UTC (0316480)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -770,8 +770,9 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 
     1. Let |observable| be a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an
        algorithm that takes a {{Subscriber}} |subscriber| and does the following:
+       1. Let |remaining| be |amount|.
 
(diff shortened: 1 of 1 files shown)


domfarolino (approved):
Great catch! Looks like the tests already assert this as well, so this is a simple bug fix that we just land right now.

Issue #127, opened 2024-03-07
Observable#from() @@iterator side effects


domfarolino:
Imagine the following:

```js
const iterable = {
  get [Symbol.iterator]() {
    console.log('Symbol.iterator getter');
    return function () {
      console.log('Symbol.iterator implementation');
      return {
        next () {
          console.log('next() being called');
          return {value: undefined, done: true};
        }
      };
    }
  }
};

const source = Observable.from(iterable);
source.subscribe();
source.subscribe();
```

What would you expect the console to look like? I think we have two options:

### First option

```
Symbol.iterator getter
Symbol.iterator implementation
next() being called
Symbol.iterator implementation
next() being called
```

This option basically means that:
 1. Observable.from() calls the `[Symbol.iterator]` getter initially to make sure `iterable` is exactly that
 2. The `[Symbol.iterator]` function is stored in the Observable
 3. It is called on each `subscribe()` invocation

Further this means that if `[Symbol.iterator]` is re-assigned in between `from()` and `subscribe()`, the _old_ value will be called on subsequent subscriptions:

```js
const iterable = {
  [Symbol.iterator]() {
    console.log('Symbol.iterator implementation');
    return {
      next () {
        console.log('next() being called');
        return {value: undefined, done: true};
      }
    };
  }
};

const source = Observable.from(iterable);
iterable[Symbol.iterator] = () => {
  throw new Error('custom error');
};

source.subscribe();
source.subscribe();
```

would output:

```
Symbol.iterator implementation
next() being called
Symbol.iterator implementation
next() being called
```

The new overridden `[Symbol.iterator]` implementation never gets called for the Observable created before the assignment.

### Second option

```
Symbol.iterator getter
Symbol.iterator getter
Symbol.iterator implementation
next() being called
Symbol.iterator getter
Symbol.iterator implementation
next() being called
```

This option basically means that:
 - Observable.from transiently tests the existence of the `Symbol.iterator` function (invoking the getter), ensuring the input object is indeed an iterable
 - Stores a reference to the input object in the Observable returned by `from()`, such that a fresh acquisition of the iterator function is grabbed on every single subscription

This allows the `[Symbol.iterator]` method that gets invoked, to change (via reassignment, for example) _across_ subscriptions to the same Observable.

[...]

Issue #125, opened 2024-03-05
Observable.from should throw primitive iterables (strings)


benlesh:
Along with how [Iterator helpers behave](https://github.com/tc39/proposal-iterator-helpers/issues/275#issuecomment-1502093384), if a user passes a `string`, even though it's an iterable, it's likely an error. We should throw a `TypeError` in that case.

```ts
Observable.from('some string'); // TypeError
```

bakkot:
I think you misread my comment (which was not all that clear, sorry). `Iterator.from` does _not_ throw when given a string. This makes it different from `flatMap`, which _does_ throw when the mapper returns a string, even though strings are iterable.

See discussion in https://github.com/tc39/proposal-iterator-helpers/issues/244..

petamoriken:
FYI: `ReadableStream.from` rejects `string` https://github.com/whatwg/webidl/pull/1397

ljharb:
If someone is intentionally passing a string into an `X.from` method, they explicitly want to convert a string into an X. I think this is the one case where an iterable string must be accepted.

flensrocker:
Can a usecase be returning a string from the catch operator?
How should I do this if Observable.from throws on a string? Or can I just return the string?
(having rxjs of("...") in mind)

domfarolino:
I'm sold on ensuring `Observable.from("string")` works the same as `Observable.from(Object("string"))` does today, by basically adding special handling like https://github.com/tc39/proposal-iterator-helpers/pull/250/files did for Iterator helpers. I'll ensure that this makes it into https://github.com/WICG/observable/pull/160, and I'll close this issue.

petamoriken:
@domfarolino

> I'm sold on ensuring `Observable.from("string")` works the same as `Observable.from(Object("string"))` does today, by basically adding special handling like https://github.com/tc39/proposal-iterator-helpers/pull/250/files did for Iterator helpers.

I think the behavior of `Observable.from(“string”)` and `Observable.from(Object(“string”))` should be clearly separated.

> Any time an iterable or async-iterable value (a value that has a `Symbol.iterator` or `Symbol.asyncIterator` method) is expected, primitives should be treated as if they were not iterable. Usually, this will mean throwing a `TypeError`. **If the user provides a primitive wrapper Object such as a String Object, however, it should be treated like any other Object**.

[...]

PR #160, opened 2024-07-21
Spec `from()` static converter

### LINKED ISSUE #159: takeUntil unclear on how to coerce type
`takeUntil`'s IDL is as follows:

```webidl
interface Observable {
  // takeUntil() can consume promises, iterables, async iterables, and other
  // observables.
  Observable takeUntil(any notifier);
}
```

However the first thing the [spec for `takeUntil(notifier)`](https://wicg.github.io/observable/#dom-observable-takeuntil) does with `notifier`, in step 2.3 is:

```
Subscribe to notifier given notifierObserver and options.
```

`Subscribe` assumes the given observer is an `Observable`. So I think there's a missing step to coerce an `any` to an `Observable`. Perhaps this is missing because `from(any value)` has yet to be specced? Additionally it looks as though Chrome's implementation just expects `takeUntil()` to recieve an `Observable`.

---

This is tracked in https://issues.chromium.org/issues/40282760 but I'm bringing it here also.

keithamus:
This makes a lot of progress towards #159 by specifying the semantics of `from()`, while also then using that algorithm in `takeUntil()`.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/160.html" title="Last updated on Jan 25, 2025, 5:48 AM UTC (57b2d19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/160/35be035...57b2d19.html" title="Last updated on Jan 25, 2025, 5:48 AM UTC (57b2d19)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -512,36 +512,41 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
 
          1. Let |nextPromise| be a {{Promise}}-or-undefined, initially undefined.
 
-         1. Let |nextRecord| be [$IteratorStepValue$](|iteratorRecord|).
+         1. Let |nextRecord| be [$IteratorNext$](|iteratorRecord|), and process it as follows:
 
-            Process |nextRecord| as follows:
+         1. If |nextRecord| is a [=throw completion=], then:
 
-            <dl class="switch">
-              <dt>If |nextRecord| is a [=throw completion=]</dt>
-              <dd>
-               <ol>
-                <li><p>[=Assert=]: |iteratorRecord|'s \[[Done]] is true.</p></li>
+            1. [=Assert=]: |iteratorRecord|'s \[[Done]] is true.
 
-                <li><p>Set |nextPromise| to [=a promise rejected with=] |nextRecord|'s \[[Value]].
-               </ol>
-              </dd>
+            1. Set |nextPromise| to [=a promise rejected with=] |nextRecord|'s \[[Value]].
 
-              <dt>If |nextRecord| is DONE</dt>
-              <dd>[=Queue a microtask=] to run |subscriber|'s {{Subscriber/complete()}} method, and
-              return from |nextAlgorithm|.
+         1. Otherwise, if |nextRecord| is [=normal completion=], then set |nextPromise| to [=a
+            promise resolved with=] |nextRecord|'s \[[Value]].
 
-              <dt>If |nextRecord| is [=normal completion=]</dt>
-              <dd>
-               <p>Set |nextPromise| to [=a promise resolved with=] |nextRecord|'s \[[Value]].</p>
-
-               Note: This is done in case |nextRecord|'s \[[Value]] is not *itself* already a {{Promise}}.
-            </dl>
+            Note: This is done in case |nextRecord|'s \[[Value]] is not *itself* already a
+            {{Promise}}.
 
          1. [=React=] to |nextPromise|:
 
-            1. If |nextPromise| was fulfilled with value |v|, then:
+            1. If |nextPromise| was fulfilled with value |iteratorResult|, then:
 
-               1. Run |subscriber|'s {{Subscriber/next()}} method, given |v|.
+               1. If [$Type$](|iteratorResult|) is not Object, then run |subscriber|'s
+                  {{Subscriber/error()}} method with a {{TypeError}} and abort these steps.
+
+               1. Let |done| be [$IteratorComplete$](iteratorResult).
+
+               1. If |done| is a [=throw completion=], then run |subscriber|'s
+                  {{Subscriber/error()}} method with a {{TypeError}} and abort these steps.
+
+                  Otherwise, if |done|'s \[[Value]] is true, then run |subscriber|'s
+                  {{Subscriber/complete()}} and abort these steps.
+
+               1. Let |value| be [$IteratorValue$](|iteratorResult).
+
+               1. If |value| is a [=throw completion=], then run |subscriber|'s
+                  {{Subscriber/error()}} method with a {{TypeError}} and abort these steps.
+
+               1. Run |subscriber|'s {{Subscriber/next()}} method, given |value|'s \[[Value]].
 
                1. Run |nextAlgorithm|, given |subscriber| and |iteratorRecord|.
 
--- a/spec.bs
+++ b/spec.bs
@@ -474,7 +474,7 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
 </div>
 
 <div algorithm>
-  To <dfn for=Observable>convert to an Observable</dfn> given an {{any}} |value|, run these steps:
+  To <dfn for=Observable>convert to an Observable</dfn> an {{any}} |value|, run these steps:
 
   Note: We split this algorithm out from the Web IDL {{Observable/from()}} method, so that
   spec prose can <a for=Observable lt="convert to an observable">convert</a> an {{Observable}}
@@ -515,17 +515,32 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
     1. Let |nextAlgorithm| be the following steps, given a {{Subscriber}} |subscriber| and an
        Iterator Record |iteratorRecord|:
 
-         1. If |iteratorRecord|'s \[[Done]] is true, then: run |subscriber|'s
-            {{Subscriber/complete()}} method and abort these steps.
+         1. Let |nextPromise| be a {{Promise}}-or-undefined, initially undefined.
 
          1. Let |nextRecord| be [$IteratorStepValue$](|iteratorRecord|).
 
-         1. Let |nextPromise| be undefined.
+            Process |nextRecord| as follows:
 
-         1. If |nextRecord| is a [=throw completion=], then set |nextPromise| to [=a promise
-            rejected with=] |nextRecord|'s \[[Value]].
+            <dl class="switch">
+              <dt>If |nextRecord| is a [=throw completion=]</dt>
+              <dd>
+               <ol>
+                <li><p>[=Assert=]: |iteratorRecord|'s \[[Done]] is true.</p></li>
 
-            Otherwise, set |nextPromise| to |nextRecord|'s \[[Value]].
+                <li><p>Set |nextPromise| to [=a promise rejected with=] |nextRecord|'s \[[Value]].
+               </ol>
+              </dd>
+
+              <dt>If |nextRecord| is DONE</dt>
+              <dd>[=Queue a microtask=] to run |subscriber|'s {{Subscriber/complete()}} method, and
+              return from |nextAlgorithm|.
+
+              <dt>If |nextRecord| is [=normal completion=]</dt>
+              <dd>
+               <p>Set |nextPromise| to [=a promise resolved with=] |nextRecord|'s \[[Value]].</p>
+
+               Note: This is done in case |nextRecord|'s \[[Value]] is not *itself* already a {{Promise}}.
+            </dl>
 
          1. [=React=] to |nextPromise|:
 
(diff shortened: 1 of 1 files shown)


domfarolino:
Thanks so much for adding this! It's been on my plate but alas I've not gotten to it yet. One thing is that we won't be able to land this without WPTs, so if you feel like writing those that would be great, otherwise I'm happy to do so, just FYI I am OOO for a bit coming up soon.

petamoriken:
The `ReadableStream.from` spec is in the process of being changed to use the newly introduced `async iterable` WebIDL type, so `Observable.from` should be similarly specified.
https://github.com/whatwg/streams/pull/1310

domfarolino:
I've started to revamp this a little bit to make sure the really subtle semantics of `[Symbol.iterator]` method access and how we deal with various values of that method's record match (a) the existing tests as well as (b) a bunch of new ones I'm adding in https://chromium-review.googlesource.com/c/chromium/src/+/5824955.

benlesh (changes requested):
Most of this looks okay, just a couple of comments.

We're also missing `ArrayLike`. Which is supported in `Array.from()` et al.

`ArrayLike` should handle any object with a `length` and numeric keys/indices.

It's probably not a big deal, and we can add it later, I suppose, but I think that `Iterator.from()` handles it? (I'll have to double check)... `Array.from()` definitely does.

domfarolino:
@benlesh Can you make a WPT PR with tests for the cases you're describing please?

benlesh:
@domfarolino :) It was in-process while you were typing that: https://github.com/web-platform-tests/wpt/pull/47874

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+  To <dfn for=Observable>convert to an Observable</dfn> given an {{any}} |value|, run these steps:
+
+  Note: We split this algorithm out from the Web IDL {{Observable/from()}} method, so that
+  spec prose can <a for=Observable lt="convert to an observable">convert</a> an {{Observable}}
+  without going through the Web IDL bindings.
+
+  Note: The resolution of value to its descrete types happens before
+  [=Observable/subscribe callback=] is called. This means mutations of values, such as adding
+  the iterable protocols to the object, will not take affect between the creation of the returned
+  observable, and when it is subscribed to.
```
domfarolino, line 486: I find this note a tiny bit confusing. We're saying both that:
 1. Narrowly-speaking, adding the iterable protocol to an already-converted object "will not take affect" (spelling). I guess this means that:
    ```js
    const promise = Promise.resolve(10);
    const o = Observable.from(promise);
    promise[Symbol.iterator] = () => {../*Produce the value 11*/};
    o.subscribe(); // Emits value 10.
    ```
 2. Broadly-speaking, "mutations of values" ... "will not take affect". This is a much stronger statement that I'm not comfortable making. In Chromium, we might end up taking the option (2) path described in https://github.com/WICG/observable/issues/127, which means the following mutation would be web-observable in between conversion and subscription:
    ```js
    const iterable = {
      [Symbol.iterator]() { /* ... */ },
    };
    const o = Observable.from(iterable);
    iterable[Symbol.iterator] = () => {throw new Error('foo');}
    o.subscribe(); // Throws error
    ```
The latter is tested in https://github.com/web-platform-tests/wpt/blob/master/dom/observable/tentative/observable-from.any.js#L164-L176 for now, and is easiest to implement in Chromium. So I think I'll either amend or remove this note entirely.

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+       Note: This step is web-observable, and re-throws any errors that the {{%Symbol.iterator%}}
+       method *getter* might have thrown.
+
+    1. If |asyncIteratorMethodRecord|'s \[[Value]] is undefined or null, then jump to the step
+       labeled <a href=#from-iterable-conversion>From iterable</a>.
+
+    1. If [$IsCallable$](|asyncIteratorMethodRecord|'s \[[Value]]) is false, then
+       [=exception/throw=] a {{TypeError}}.
+
+       Otherwise, let |nextAlgorithm| be the following steps, given |subscriber| and |iterator|:
```
domfarolino, line 515: I've cleaned up the async iterator conversion logic a good bit, but I think one thing we're still not doing is checking for DONE-ness. I think at some point we need to be checking if the iterable is DONE, i.e., a line like https://streams.spec.whatwg.org/#readable-stream-from-iterable:~:text=Let%20done%20be%20%3F%20IteratorComplete(iterResult).

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+
+    1. If [$IsCallable$](|iteratorMethodRecord|'s \[[Value]]) is false, then [=exception/throw=] a
+       {{TypeError}}.
+
+       Otherwise, return a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an
+       algorithm that takes a {{Subscriber}} |subscriber| and does the following:
+
+         1. Let |iteratorRecord| be [$GetIterator$](|value|, sync).
+
+            Note: This re-invokes any method getters for the {{%Symbol.iterator%}} method on
```
domfarolino, line 572: This is what I was referring to earlier when I said _some_ mutations to the underlying object (in between conversion and subscription) will take effect, and be web-observable. (Same goes for async iteratbles).

domfarolino (commented):
Quick review just to document the current state of things. I think we're just about ready modulo the async iterator stuff, which I think needs a little work I've outlined below.

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+
+               1. Run |nextAlgorithm|, given |subscriber| and |iterator|.
+
+            1. If |nextPromise| was rejected with reason |r|, then run |subscriber|'s
+               {{Subscriber/error()}} method, given |r|.
+
+    1. Return a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an algorithm that
+       takes a {{Subscriber}} |subscriber| and does the following:
+
+       1. Let |iteratorRecord| be [$GetIteratorFromMethod$](|value|, {{%Symbol.asyncIterator%}}).
```
domfarolino, line 543: This declaration is wrong, `GetIteratorFromMethod()` requires an actual instance method, not the well-known name. In general I think this can just use `GetIterator(|value|, async)`, so I'll change it to that.

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+
+       1. If |iteratorRecord| is a [=throw completion=], then [=queue a microtask=] to run
+          |subscriber|'s {{Subscriber/error()}} method, given |iteratorRecord|'s \[[Value]].
+
+       1. Otherwise, [=queue a microtask=] to run |nextAlgorithm| given |subscriber| and
+          |iteratorRecord|'s \[[Value]].
```
domfarolino, line 549: `|iteratorRecord|` does not have a `[[Value]]` field if we've made it this far. `GetIteratorFromMethod()` (and `GetIterator()` for that matter) [both return](https://tc39.es/ecma262/#sec-getiteratorfrommethod) an [Iterator Record](https://tc39.es/ecma262/#sec-iterator-records) which has a `[[Iterator]]`, `[[NextMethod]]` and `[[Done]]`, but no `[[Value]]`. Maybe we should pass the full record in here.

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+
+    1. If [$IsCallable$](|asyncIteratorMethodRecord|'s \[[Value]]) is false, then
+       [=exception/throw=] a {{TypeError}}.
+
+       Otherwise, let |nextAlgorithm| be the following steps, given |subscriber| and |iterator|:
+
+         1. If |iterator|'s \[[Done]] is true, then: run |subscriber|'s {{Subscriber/complete()}}
+            method and abort these steps.
+
+         1. Let |nextRecord| be [$IteratorStepValue$](|iterator|).
```
domfarolino, line 520: First off, we might want to use `IteratorNext()` instead of `IteratorStepValue()`. I'll investigate this. Second, we need to handle throw completions here, since we're not doing that currently. Probably we want to queue a microtask to call `error()` on abrupt completions, with the completion value.

spec.bs
```diff
@@ -461,6 +473,146 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+
+    1. Return a [=new=] {{Observable}} whose [=Observable/subscribe callback=] is an algorithm that
+       takes a {{Subscriber}} |subscriber| and does the following:
+
+       1. Let |iteratorRecord| be [$GetIteratorFromMethod$](|value|, {{%Symbol.asyncIterator%}}).
+
+       1. If |iteratorRecord| is a [=throw completion=], then [=queue a microtask=] to run
+          |subscriber|'s {{Subscriber/error()}} method, given |iteratorRecord|'s \[[Value]].
+
+       1. Otherwise, [=queue a microtask=] to run |nextAlgorithm| given |subscriber| and
```
domfarolino, line 548: I don't think we want to queue this microtask here. I understood the original intent (so that in the event the iterator is `[[Done]]`, we run `complete()` asynchronously), but from looking at this more the IteratorRecord that `GetIterator()` always returns has `[[Done]]: false`, so I think we don't need to do this up-front checking at all. Furthermore, queuing this microtask has web-observable consequences.

With it, the invocation of `next()` is behind at least two microtasks. First, is the one we queue here, and second is the one is the one we get by "reacting" to the promise returned from the iterator's `[[NextMethod]]`. I think this is web-observable, because if the retrieval and/or invocation of `%Symbol.asyncIterator%` itself queues a microtask, this microtask would interleave the two aforementioned ones, which seems weird.

To simplify the structure of this, I think we should move the done check and `complete()` call after the `next()` method invocation (behind the reaction microtask, but still synchronous with respect to `subscriber->next()`). That way it's still queued behind a microtask, but it's in the right place.

spec.bs
domfarolino: I missed it, but I guess we are doing that in the first step of `nextAlgorithm`. But I think the handling of this is wrong anyways. See the other comment below about moving the done check.

domfarolino (commented):
A few more things I've noted. I'll work on these and try and fix most of them today.

spec.bs
domfarolino: Fixed in https://github.com/WICG/observable/pull/160/commits/4c74aa8b0ab8046fe996855b0bcd69e941c1e6da.

spec.bs
domfarolino: Done in https://github.com/WICG/observable/pull/160/commits/d328e70c695e44a8f915971ac71580c2c78fb4ea.

spec.bs
domfarolino: Done in https://github.com/WICG/observable/pull/160/commits/4a34553cc13fe1e7b06c33db5b02e8051dca71bf. I've left `IteratorStepValue()` alone here since that seems fine, but I've handled throw completions and the Done value _after_ we pull the next value, since `[[Done]]` should always be false the very first time, so we don't need to check it before we step forward.

spec.bs
domfarolino: Done in https://github.com/WICG/observable/pull/160/commits/4a34553cc13fe1e7b06c33db5b02e8051dca71bf. Most of the above comment was right I think, but we don't want to move the done check to after we invoke `subscriber->next()`. I've restructured things a bit and added a note about microtask timing.

spec.bs
domfarolino: Removed in https://github.com/WICG/observable/pull/160/commits/06a1a66575808372a57a83be737af4b8aa61e025.

spec.bs
domfarolino: Actually I think we can't use `IteratorStepValue()`, since it expects the `next()` method to return an `IteratorResult` object with a `value` property. But for async iterators, per https://tc39.es/ecma262/#table-async-iterator-required, the `next()` method is expected to return a `Promise` (and if it isn't a `Promise` we wrap it in anyways), which resolves _to_ an `IteratorResult`. So I think we do have to use `IteratorNext()` like the Streams Standard does, and react to the returned Promise, and inside the reaction branch can we grab the value from the object etc.

domfarolino: Done in https://github.com/WICG/observable/pull/160/commits/1b92dc5c14c6d80adb6a55e16231c1072ec7393d.

spec.bs
```diff
@@ -461,6 +473,165 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
...
+  Note: We split this algorithm out from the Web IDL {{Observable/from()}} method, so that
+  spec prose can <a for=Observable lt="convert to an observable">convert</a> values to without
+  going through the Web IDL bindings.
+
+    1. If [$Type$](|value|) is not [=Object=], [=exception/throw=] a {{TypeError}}.
+
+       Note: This prevents primitive types from being coerced into iterables (e.g., String).
+
+       Issue: See if this is even the behavior we want. See <a
+       href=https://github.com/WICG/observable/issues/125>WICG/observable#125</a>
```
domfarolino, line 488: See https://github.com/WICG/observable/issues/125#issuecomment-2486600544 where I think we want to reverse this, perhaps requiring special handling to ensure that Strings can be converted into Observables like an Iterable would be.

benlesh: Nah... `Observable.from(variable)` where `variable` is a string, is almost always a mistake. I think that `Iterator.from` et al, don't allow `string`. I think the wording is "non-primitive iterables" or some such thing. I'd have to research. But we don't want to allow string, because there's just not a great use case for it, IMO. They could still do it like `Observable.from(someString[Symbol.iterator]())` if they REALLY wanted to.

domfarolino: > I think that `Iterator.from` et al, don't allow `string`

No, see https://github.com/WICG/observable/issues/125#issuecomment-1979742835. Regardless, we could start by throwing in this case and eventually relax it if we want, but it seems like for consistency it would be weird if Iterables convert strings properly, Observables convert Iterables properly, but Observables do not convert strings like Iterables.

domfarolino (approved):
Replying to https://github.com/WICG/observable/pull/160#pullrequestreview-2269846677 about `ArrayLike`: based on https://github.com/tc39/proposal-iterator-helpers/issues/231 it seems like `Iterator.from()` very intentionally does NOT support `ArrayLike`, and based on the guidance in that issue it seems like we should not either.

The latest commit https://github.com/WICG/observable/pull/160/commits/57b2d19e6fb4415d1d25bb70ce35a3c67a2bcee4 removes the async iterator conversion stuff here, since it got a little complicated and I'd like to focus on it separately from this PR; I'll re-add it, clean it up, and land it as a follow-up to this PR. I've filed https://github.com/WICG/observable/issues/191 to track the two remaining comments in this PR regarding the async iterable conversion semantics, to make sure they get addressed in the follow-up.

Given that, I'll proceed to approve this PR and land it, and follow-up with the async iterable conversion work.

PR #163, opened 2024-07-26
"subscribe to an": use |observer| instead of [=this=]

keithamus:
The "subscribe to an" algorithm references `[=this=]` which I believe is likely due to it being refactored from a method to an independant algorithm. It should use `|observable|`, referencing the given observable, which is otherwise not mentioned in the algorithm.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/163.html" title="Last updated on Jul 26, 2024, 1:10 PM UTC (6e1697a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/163/1b470a5...6e1697a.html" title="Last updated on Jul 26, 2024, 1:10 PM UTC (6e1697a)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -455,7 +455,7 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
   bindings on objects whose properties could be mutated by JavaScript. See
   [[#promise-returning-operators]] for usage of this.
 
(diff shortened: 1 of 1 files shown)


keithamus:
Oh I'm mistaken here. `[=this=]` is actually proper here.

PR #166, opened 2024-07-27
Pass abortsignal reason to close a subscription

### LINKED ISSUE #165: Closing subscriptions with AbortSignal reasons
The WPTs asser that an AbortSignal's reason will be passed through when a subscription is closed, for example [when subscribing with an already aborted signal](https://github.com/web-platform-tests/wpt/blob/ad7e196b122827421b84dca83cda6d69e64eef41/dom/observable/tentative/observable-constructor.any.js#L306). The problem is the spec doesn't seem to provide those mechanics. Qoth the spec:

> To subscribe to an [Observable](https://wicg.github.io/observable/#observable)
>
> ...
>
> 6. If options’s [signal](https://wicg.github.io/observable/#dom-subscribeoptions-signal) [exists](https://infra.spec.whatwg.org/#map-exists), then:
>    1. If options’s [signal](https://wicg.github.io/observable/#dom-subscribeoptions-signal) is [aborted](https://dom.spec.whatwg.org/#abortsignal-aborted), then [close](https://wicg.github.io/observable/#close-a-subscription) subscriber.
>
> ...
>

The problem here is that then ["close a subscription"](https://wicg.github.io/observable/#close-a-subscription) isn't passed the `reason` and so doesn't get to abort its own signal with that reason.

I think, based on the WPTs, it seems as though we need to update the "close a subscription" step to have an optional reason so it can propagate.

keithamus:
Fixes #165. 

This seems to marry up with [Chrome's impl](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/subscriber.cc;l=64-72;drc=105770df485ace262780d95126bb60b1a16ec340;bpv=0;bpt=1):

```c++
    if (downstream_signal->aborted()) {
      CloseSubscription(
          script_state,
          /*abort_reason=*/downstream_signal->reason(script_state));
    } else {
      close_subscription_algorithm_handle_ = downstream_signal->AddAlgorithm(
          MakeGarbageCollected<CloseSubscriptionAlgorithm>(
              this, downstream_signal, script_state));
    }
```

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/166.html" title="Last updated on Aug 13, 2024, 5:08 PM UTC (7905c52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/166/530d498...7905c52.html" title="Last updated on Aug 13, 2024, 5:08 PM UTC (7905c52)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -37,7 +37,7 @@ urlPrefix: https://dom.spec.whatwg.org; spec: DOM
     for: AbortSignal
       text: dependent signals; url: abortsignal-dependent-signals
       text: signal abort; url:abortsignal-signal-abort
-      text: reason; url:abortsignal-abort-reason
+      text: abort reason; url:abortsignal-abort-reason
 </pre>
 
 <style>
@@ -531,13 +531,13 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
 
        1. If |options|'s {{SubscribeOptions/signal}} is [=AbortSignal/aborted=], then [=close a
           subscription|close=] |subscriber| given |options|'s {{SubscribeOptions/signal}}
-          [=AbortSignal/reason=].
+          [=AbortSignal/abort reason=].
 
        1. Otherwise, [=AbortSignal/add|add the following abort algorithm=] to |options|'s
           {{SubscribeOptions/signal}}:
 
           1. [=close a subscription|Close=] |subscriber| with |options|'s
-             {{SubscribeOptions/signal}} [=AbortSignal/reason=].
+             {{SubscribeOptions/signal}} [=AbortSignal/abort reason=].
 
     1. If [=this=]'s [=Observable/subscribe callback=] is a {{SubscribeCallback}}, [=invoke=] it
        with |subscriber|.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -37,6 +37,7 @@ urlPrefix: https://dom.spec.whatwg.org; spec: DOM
     for: AbortSignal
       text: dependent signals; url: abortsignal-dependent-signals
       text: signal abort; url:abortsignal-signal-abort
+      text: reason; url:abortsignal-abort-reason
```
domfarolino, line 40: Just to double check, is this needed? I guess it is, but I'm surprised by this. It makes me wonder if we're holding AbortSignal wrong — I would expect this to be exported so other specs that use AbortSignals can propagate reasons similarly.

spec.bs
```diff
@@ -292,9 +294,10 @@ observable.subscribe({}, {signal: outerController.signal});
 
     1. Set |subscriber|'s [=Subscriber/active=] boolean to false.
 
-    1. [=AbortSignal/Signal abort=] |subscriber|'s [=Subscriber/subscription controller=].
+    1. [=AbortSignal/Signal abort=] |subscriber|'s [=Subscriber/subscription controller=]
+       with |reason|, if it is given.
 
-       Issue: Abort with an appropriate abort reason.
+       Issue: Abort with an appropriate abort reason if none given.
```
domfarolino, line 300: I don't think this is needed at all now. We don't need to supply a "fallback" default reason if none was provided.

domfarolino (approved):

domfarolino:
I cleaned a few things up here. I think it's good to go now so I'll merge.

PR #167, opened 2024-08-02
Let the `value` argument be consistent

maraisr:
Am currently implementing `forEach` in WebKit, and noticed we had some [inconsistencies](https://github.com/WebKit/WebKit/blob/d128fcbc4cf949b84f671691d19297f0fda88349/Source/WebCore/dom/MapperCallback.idl#L26) with what we call the value we pass these callbacks. Lets tidy this up in the spec.

cc @keithamus

***
<a href="https://pr-preview.s3.amazonaws.com/maraisr/observable/pull/167.html" title="Last updated on Aug 2, 2024, 3:41 AM UTC (ec10807)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/167/530d498...maraisr:ec10807.html" title="Last updated on Aug 2, 2024, 3:41 AM UTC (ec10807)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -333,10 +333,10 @@ dictionary SubscribeOptions {
 
 callback Predicate = boolean (any value, unsigned long long index);
 callback Reducer = any (any accumulator, any currentValue);
-callback Mapper = any (any element, unsigned long long index);
+callback Mapper = any (any value, unsigned long long index);
 // Differs from Mapper only in return type, since this callback is exclusively
 // used to visit each element in a sequence, not transform it.
-callback Visitor = undefined (any element, unsigned long long index);
+callback Visitor = undefined (any value, unsigned long long index);
 
 [Exposed=*]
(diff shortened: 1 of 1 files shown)


keithamus (approved):
I think this is a sensible change given that `Predicate` takes a `value`, and also `element` _may_ be confusing for developers in the context of the DOM.

domfarolino:
Thanks so much! This makes sense. Happy to approve and bulldoze the IPR bot this time, but if you plan on making more (larger, is the key) contributions, then I think we should have you sign up for WICG so that the bot is by default happy. Feel free to sign up with the link in the checks section above.

domfarolino (approved):

maraisr:
Ah thanks for that @domfarolino ❤️ I do have a W3C account, and have that affiliation set up. Did I need to link something to see that reflected here?

PR #171, opened 2024-08-27
Spec the `reduce()` operator

domfarolino:
This specs the `reduce()` operator discussed in https://github.com/WICG/observable/issues/126. Tests will be upstreamed in https://chromium-review.googlesource.com/c/chromium/src/+/5817639. That should leave only `catch()` and `finally()` operators left.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/171.html" title="Last updated on Aug 28, 2024, 2:19 AM UTC (2760e00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/171/69ad0de...2760e00.html" title="Last updated on Aug 28, 2024, 2:19 AM UTC (2760e00)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1628,17 +1628,19 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
 
     1. Let |idx| be an {{unsigned long long}}, initially 0.
 
-    1. Let |accumulator| be |initialValue| if it is given, and the string "<code>unset</code>" otherwise.
+    1. Let |accumulator| be |initialValue| if it is given, and the string "<code>unset</code>"
+       otherwise.
 
     1. Let |observer| be a new [=internal observer=], initialized as follows:
 
        : [=internal observer/next steps=]
-       :: 1. If |accumulator| is the string "<code>unset</code>", then set |accumulator| to the
-             passed in |value| and abort these steps.
+       ::
+          1. If |accumulator| is the string "<code>unset</code>", then set |accumulator| to the
+             passed in |value|, set |idx| to |idx| + 1, and abort these steps.
 
              Note: This means that |reducer| will not be called with the first |value| that [=this=]
              produces set as the {{Reducer/currentValue}}. Rather, when the *second* value is
-             eventually emitted, we will call |reducer| with *it* as the {[Reducer/currentValue}},
+             eventually emitted, we will call |reducer| with *it* as the {{Reducer/currentValue}},
              and the first value (that we're saving here) as the {{Reducer/accumulator}}.
 
           1. [=Invoke=] |reducer| with |accumulator| as {{Reducer/accumulator}}, the passed in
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1604,7 +1604,68 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
...
+
+    1. [=AbortSignal/add|Add the following abort algorithm=] to |internal options|'s
+       {{SubscribeOptions/signal}}:
+
+       1. [=Reject=] |p| with |internal options|'s {{SubscribeOptions/signal}}'s [=AbortSignal/abort
+          reason=].
+
+    1. Let |idx| be an {{unsigned long long}}, initially 0.
+
+    1. Let |accumulator| be |initialValue| if it is given, and the string "<code>unset</code>"
```
benlesh, line 1631: Is this how the Array reduce internals are spec'ed? It seems mildly inefficient.

Most implementations I've seen will either use a `hasValue` boolean, or check the current `index` and whether or not an `initialValue` was provided to determine whether or not to call the reducer.

Jamesernator, line 1631: Using a string here is a hazard with user code as this string could well be passed as the initial value by user code, or returned from the reducer.

domfarolino: Ah yeah, for sure. As I was implementing this "unset" was really just an internal-only value, so I'll ensure that the alternate form of this variable is something like that too, instead of something that developers can literally pass in heh. Whoops.

domfarolino: Checking `hasValue` each time a new value is emitted from the Observable will definitely be as inefficient as checking if accumulator is "unset" or some other special internal value each time, since the check is being performed _each time_ regardless, right? So I'm not sure I see an efficiency difference here. But regarding the comment in https://github.com/WICG/observable/pull/171#pullrequestreview-2264670568, I will be migrating to something like `hasValue` anyways.

domfarolino: Also FWIW, specs are not concerned with efficiency as much as implementations. They are concerned with whatever semantics are observable to script, and engines are free to optimize things as much as they want or can, provided they have the same observable semantics as specifications.

domfarolino: I've decided to use the uninitialized state of `|accumulator|` as a signal for whether `|initialValue|` was passed in initially. That saves us spec-wise from having to maintain a separate variable of `|accumulatorHasValue|` or something like that, yet has the same semantics.

benlesh (approved):
LGTM. As discussed offline, there's the obvious issue that there's not an analog for `observable.reduce(fn)` that takes a cancellation, because of the argument order: `observable.reduce(fn, undefined, { signal })`. But that doesn't seem like a big deal. If it ever IS a big deal, there could be an option added to ignore the initialValue, I suppose. But that's one of those "why is this there?" sort of things. :)

PR #172, opened 2024-08-28
Spec the `catch()` operator

### LINKED ISSUE #64: I think `catch` may be a requirement

domfarolino:
Tests are being upstreamed in https://chromium-review.googlesource.com/c/chromium/src/+/5823731.

Closes #64.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/172.html" title="Last updated on Aug 28, 2024, 4:07 PM UTC (761aee1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/172/69ad0de...761aee1.html" title="Last updated on Aug 28, 2024, 4:07 PM UTC (761aee1)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -351,6 +351,10 @@ callback Mapper = any (any value, unsigned long long index);
 // used to visit each element in a sequence, not transform it.
 callback Visitor = undefined (any value, unsigned long long index);
 
+// This callback returns an `any` that must convert into an `Observable`, via
+// the `Observable` conversion semantics.
+callback CatchCallback = any (any value);
+
 [Exposed=*]
 interface Observable {
   constructor(SubscribeCallback callback);
@@ -375,6 +379,7 @@ interface Observable {
   Observable flatMap(Mapper mapper);
   Observable switchMap(Mapper mapper);
   Observable inspect(optional ObservableInspectorUnion inspect_observer = {});
+  Observable catch(CatchCallback callback);
   Observable finally(VoidFunction callback);
 
   // Promise-returning operators.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1207,6 +1212,69 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
...
+
+       1. Let |sourceObserver| be a new [=internal observer=], initialized as follows:
+
+          : [=internal observer/next steps=]
+          :: Run |subscriber|'s {{Subscriber/next()}} method, given the passed in <var
+             ignore>value</var>.
+
+          : [=internal observer/error steps=]
+          :: 
+             1. [=Invoke=] |callback| with the passed in <var ignore>error</var>. Let |result| be
```
benlesh, line 1231: I don't know if it's worth mentioning in a note here, but because the source observable errored, it's already finalized. Otherwise, it almost looks like we're missing the step of unsubscribing from the source here.

benlesh (approved):
Approved with a note.

spec.bs
domfarolino: More documentation can't hurt. I've added a note just below where we subscribe to `|innerObservable|` since that's the part that might look the funniest without context.

PR #179, opened 2024-10-16
Update spec to mention rejecting with a new RangeError

maraisr:
Reading the [Chrome Implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/observable.cc;l=577-581?q=Observables%20foreach&ss=chromium) of `.first`, and the associated [wpt tests](https://github.com/web-platform-tests/wpt/blob/3585305e8365779fac42dd87e4f52c89c3e884a6/dom/observable/tentative/observable-first.any.js#L38-L54) I think it is valuable in having the spec reflective of this.

***
<a href="https://pr-preview.s3.amazonaws.com/maraisr/observable/pull/179.html" title="Last updated on Oct 20, 2024, 11:29 AM UTC (f77730e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/179/a56ef28...maraisr:f77730e.html" title="Last updated on Oct 20, 2024, 11:29 AM UTC (f77730e)">Diff</a>

PR #180, opened 2024-10-21
Observable#last should also reject with RangeError with no last value

maraisr:
With the last change in https://github.com/WICG/observable/pull/179, for observable first — we probably want the same to be true for last. There is an inflight wpt update diff for this too: https://github.com/web-platform-tests/wpt/pull/48717

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1243,7 +1243,7 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
              ignore>value</var>.
 
           : [=internal observer/error steps=]
-          :: 
+          ::
              1. [=Invoke=] |callback| with the passed in <var ignore>error</var>. Let |result| be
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1243,7 +1243,7 @@ For now, see [https://github.com/wicg/observable#operators](https://github.com/w
              ignore>value</var>.
 
           : [=internal observer/error steps=]
-          :: 
+          ::
```
maraisr, line 1246: My IDE auto trimmed this whitespace, arguably unrelated to this change — so happy to revert if so.

maraisr:
cc @domfarolino

keithamus (approved):

domfarolino:
(Thanks for taking care of this, and thanks @keithamus for getting to it so quickly).

PR #183, opened 2024-11-12
Correct example in README.md

### LINKED ISSUE #184: Error in websocket example in README.md
The `subscribe` function doesn't return a subscription, so the return value should not be stored in a local variable.

See #183

flensrocker:
In the websocket example the return value of the subscribe calls are stored in variables.

The current spec doesn't implement a "subscription" return value, this might be a left over from the past.

Fixes #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ const googTrades = streamStock('GOOG');
 const nflxTrades = streamStock('NFLX');
 
 const googController = new AbortController();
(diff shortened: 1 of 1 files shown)


domfarolino (approved):
Thanks a lot!

domfarolino:
The build is failing for an unrelated reason that I'll figured out separately from this PR, so I'll just land this.

PR #187, opened 2024-12-11
We should forward the subscribers abort signal to eventListeners

maraisr:
I think there is a possible event listener leak that can occur if we don't have a mechanism to clean up event listeners as subscriptions are aborted.

In WebKit I confirmed, that if we were to follow the spec we'd keep listeners around for a little too long:

CleanShot 2024-12-11 at 09 32 01

Much [like Chrome's EventTarget implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/events/event_target.cc;l=362-366), WebKits' also immediately take ownership of the EventListener, so that despite the Observable is marked for GC (`active == 0`), the listener hangs around.

and I think Chrome had a similar discovery, as they [also forward the signal](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/events/event_target.cc;l=271;drc=616d60fca655937c2b730db94fd32d37ddff3bb5) to the event listener to trigger the abort steps.

***
<a href="https://pr-preview.s3.amazonaws.com/maraisr/observable/pull/187.html" title="Last updated on Dec 12, 2024, 5:22 AM UTC (391600b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/187/a8bfae2...maraisr:391600b.html" title="Last updated on Dec 12, 2024, 5:22 AM UTC (391600b)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1811,11 +1811,11 @@ partial interface EventTarget {
                :: false
 
                : [=event listener/signal=]
-               :: null
+               :: |subscriber|'s [=Subscriber/subscription controller=]'s [=AbortController/signal=]
 
-                  Note: The {{AbortSignal}} for event listeners [=add an event listener|added=] by
-                  {{EventTarget/when()}} is managed by the {{Observable}} itself. See
-                  {{Observable/subscribe()}} and {{SubscribeOptions}}.
+                  Note: This ensures that the [=event listener=] is cleaned up when
+                  [=Subscriber/subscription controller=]'s [=AbortController/signal=] is [=AbortSignal/aborted=],
+                  regardless of an engine's ownership model.
 
     1. Return |observable|.
 </div>


maraisr:
cc @domfarolino

spec.bs
```diff
@@ -1811,11 +1811,12 @@ partial interface EventTarget {
                :: false
 
                : [=event listener/signal=]
-               :: null
+               :: |subscriber|'s [=Subscriber/subscription controller=]'s [=AbortController/signal=]
 
-                  Note: The {{AbortSignal}} for event listeners [=add an event listener|added=] by
-                  {{EventTarget/when()}} is managed by the {{Observable}} itself. See
-                  {{Observable/subscribe()}} and {{SubscribeOptions}}.
+                  Note: To ensure that the [=event listener=] is cleaned up when
```
domfarolino, line 1816:
```suggestion
                  Note: This ensures that the [=event listener=] is cleaned up when
```

spec.bs
```diff
@@ -1811,11 +1811,12 @@ partial interface EventTarget {
...
+                  [=Subscriber/subscription controller=]'s [=AbortController/signal=] is [=AbortSignal/aborted=].
```
domfarolino, line 1817: I would add ", regardless of an engine's ownership model". And remove the last sentence. I think that makes the most sense.

domfarolino (commented):
Thanks! This is a great catch.

maraisr:
Thank you so much for your review here @domfarolino, I have addressed the points. You're so good at this spec _stuff_. Learning a bunch.

domfarolino (approved):
Thanks a lot for the contribution!

domfarolino:
Despite me updating https://github.com/WICG/observable/commit/a8bfae276265dc2d29e6009f7a6003da280df160, this is still failing for a pesky build reason. Would you be able to rebase this PR off of the new master branch (on top of https://github.com/WICG/observable/commit/a8bfae276265dc2d29e6009f7a6003da280df160 basically) and we can try the build again? I think it might be building with an old version of ubuntu.

maraisr:
@domfarolino I have rebased from the current master. Let's hope that fixes the CI.

PR #194, opened 2025-01-29
Add some words about async iteration

benlesh:
This is meant to address questions about why we don't want to use `AsyncIterator` to handle events.

cc @domfarolino

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,17 @@ await element.when('mousemove')
 ```js
 // Imperative
 const controller = new AbortController();
-element.addEventListener('mousemove', e => {
-  console.log(e);
+element.addEventListener(
+	'mousemove',
+	(e) => {
+		console.log(e);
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,13 +51,17 @@ await element.when('mousemove')
...
+element.addEventListener(
+	'mousemove',
+	(e) => {
+		console.log(e);
+
+		element.addEventListener('mouseup', (e) => {
+			controller.abort();
+		});
+	},
+	{ signal: controller.signal },
```
domfarolino, line 63: Let's revert changes like this which are just stylistic and probably automated by an editor.

benlesh: They're from our committed prettier config.

domfarolino: Can you revert them or only format the changes relevant to this PR instead of the whole file? That'd make it easier to review and better scoped.

PR #197, opened 2025-02-06
Spec ref-counted producers

### LINKED ISSUE #170: Reconsider unpredictable hot/cold and async/sync behavior
The rxjs inspired API creates a situation where Observables sometimes have side effects upon subscription (hot vs cold) and sometimes emit synchronously and sometimes async. If you have a function that takes an Observable as an argument you have no predictability if it will have a side effect upon subscription (is it safe to call it twice?), and no predictability if the emit will happen immediately when subscribing or later in a microtask.

This is a manifestation of Releasing Zalgo:
https://blog.izs.me/2013/08/designing-apis-for-asynchrony/
https://blog.ometer.com/2011/07/24/callbacks-synchronous-and-asynchronous/

Promises are the opposite of this:
* They always resolve async
* Calling an operator on a Promise never has a side effect
* Calling an operator on a Promise is always cheap

This means if someone passes a Promise to you, all operations on it are predictable and safe.

With Observables the situation is different:
* They may resolve sync or async
* Calling subscribe might be expensive (API request) or cheap (replay)

In large systems built on Observables I see this manifest often as:
* Making accidental repeated API requests (ex. calling subscribe twice and forgetting a replay)
* Adding repeated listeners (ex. many MutationObservers instead of one)
* API requests at unpredictable and undesirable times (too late when the UI is rendering or too early during critical loading)
* Confusing ordering of operations
* Dangerous refactoring where it switches from sync to async or vice versa

Other Web APIs are also consistent today (ex. MutationObserver, ResizeObserver, Promises), and never emit during subscription and have a split between observing the operation and executing it. The one place I'm aware where this was violated is MessagePort#onmessage starting the port, but it's generally been considered a mistake.

Context: I lead possibly the largest codebase built on rxjs in the world.

### LINKED ISSUE #178: Ref Counted Observable discussion
After TPAC last week, it was determined that the ideal type of observable might be a ref-counted observable. The ref-counted observable will function roughly as follows:

1. The observable is "cold" until the first subscriber.
2. When the first subscriber subscribes, the subscription logic is executed.
3. Every additional subscriber will share that subscription (it's "hot").
4. When all subscribers abort their subscriptions, the ref-count drops to 0, and teardown/unsubscription logic is executed, returning the observable to a "cold" state.

The idea behind this type is to address concerns from one issue (#170), where an RxJS user thought it was confusing that sometimes observables have side effects, and sometimes they didn't. This would guarantee that for N subscribers, there would be at most _one_ side effect.

domfarolino:
Closes #170. Closes #178.

Corresponding Chromium change _and_ tests: https://chromium-review.googlesource.com/c/chromium/src/+/6221901.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/observable/pull/197.html" title="Last updated on Feb 21, 2025, 2:09 AM UTC (a819eeb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/observable/197/df571e8...a819eeb.html" title="Last updated on Feb 21, 2025, 2:09 AM UTC (a819eeb)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -771,7 +771,7 @@ An <dfn>internal observer</dfn> is a [=struct=] with the following [=struct/item
        1. Return.
 
     1. Let |subscriber| be a [=new=] {{Subscriber}}, initialized with [=Subscriber/owning
-       Observable=] [=this=].
+       Observable=] as [=this=].
 
     1. [=set/Append=] |internal observer| to |subscriber|'s [=Subscriber/internal observers=].
 
--- a/spec.bs
+++ b/spec.bs
@@ -177,7 +177,8 @@ interface Subscriber {
 };
 </xmp>
 
-Each {{Subscriber}} has an [=ordered set=] of <dfn for=Subscriber>internal observers</dfn>, initially empty.
+Each {{Subscriber}} has an [=ordered set=] of <dfn for=Subscriber>internal observers</dfn>,
+initially empty.
 
 Each {{Subscriber}} has a <dfn for=Subscriber>teardown callbacks</dfn>, which is a [=list=] of
 {{VoidFunction}}s, initially empty.
@@ -187,7 +188,8 @@ Each {{Subscriber}} has a <dfn for=Subscriber>subscription controller</dfn>, whi
 
 Each {{Subscriber}} has a <dfn for=Subscriber>active</dfn> boolean, initially true.
 
-Each {{Subscriber}} has an <dfn for=Subscriber>owning Observable</dfn>, an {{Observable}}, which is initialized during construction.
+Each {{Subscriber}} has an <dfn for=Subscriber>owning Observable</dfn>, an {{Observable}}, which is
+initialized during construction.
 
 Note: This is a bookkeeping variable to ensure that a {{Subscriber}} never calls any of the
 callbacks it owns after it has been [=close a subscription|closed=].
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -188,6 +188,9 @@ Each {{Subscriber}} has a <dfn for=Subscriber>subscription controller</dfn>, whi
 
 Each {{Subscriber}} has a <dfn for=Subscriber>active</dfn> boolean, initially true.
 
+Each {{Subscriber}} has an <dfn for=Subscriber>owning Observable</dfn>, an {{Observable}}, which is
```
benlesh, line 191: It's actually not necessary for a Subscriber to hold a reference to an Observable. But it is necessary for a (multicast) Observable to hold a reference to the Subscriber, so this is probably okay.

This just caught my eye because with classic "cold" observables, the Observable instance itself could theoretically be GC'ed while the subscription is still alive.

domfarolino: I guess you're saying the Subscriber doesn't need to tell the Observable when it is inactive/cleared, and instead the Observable, upon new subscriptions, can just detect if its "active subscriber" is no longer active, and needs to create a new one instead?

domfarolino: > The subscriber would probably remove itself as part of its teardown steps, yeah.

That requires a Subscriber->Observable reference though, right? That conflicts with what you said about it being weird to have that reference.

Basically I think it should be possible for Subscriber to not have to have a reference to the Observable it is associated with (so that the Observable can still be GC'd without the Subscriber being killed). Do you agree? In that case when the Subscriber becomes inactive, it just marks itself as inactive, but doesn't touch the Observable at all. The Observable would be responsible for checking its associated Subscriber's "active" status upon future subscriptions. What do you think?

domfarolino: As discussed offline:

1. Subscriber does not need to reference Observable at all (upon becoming inactive, it can just set its internal state to inactive, and rely on its associated Observable checking this)
2. Observable doesn't need to strongly reference/_own_ Subscriber. Rather it can just weakly reference its subscriber. Upon subscription, the Observable just checks if (a) the weak subscriber is still alive, and if so, if it's active then add the observer to the active subscriber's collection of observers; otherwise, discard it and start a new subscription (which means running the subscription callback again etc.)

I've made these changes in https://github.com/WICG/observable/pull/197/commits/a819eeb590757e8800d00c07bdf9cd5c163f846e, so I'll resolve this and land the PR.

PR #201, opened 2025-02-25
Very important bug fix to the Konami code.

mikewest:
I smiled when I saw this in the explainer. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,8 +223,6 @@ const pattern = [
   'ArrowRight',
   'b',
   'a',
-  'b',
-  'a',
   'Enter',
 ];
 


ljharb:
this isn't a bugfix, this is a bug - there's definitely always been two "b a" sequences

mikewest:
That doesn't match my muscle memory (though I do generally remember hitting "select" before "start" so my friend and I could _both_ fail to get very far in Contra. :) ). Wikipedia suggests the same: https://en.wikipedia.org/wiki/Konami_Code.

ljharb:
huh; i could have sworn there were two, but googling suggests that was a misconception that was frequently propagated.

